PR #2, opened 2018-03-10
Fix SQL injection in advanced title search

rillig:
Currently, searching for double quotes in the advanced search leads to a
500 Internal Server Error. After this fix is applied, the same search
request should return the intended result, which is all titles that
actually contain these quotes.

This commit has not been tested, so it is rather a suggestion.

### DIFF
--- a/application/libraries/Librivox_search.php
+++ b/application/libraries/Librivox_search.php
@@ -35,9 +35,9 @@ class Librivox_search{
 		$title = ' ';
 		if (!empty($params['title']) )
 		{
-			$params['title'] = str_replace(' ', '%', $params['title']);
-			$title = 'AND st.`search_field` LIKE "' . $params['like_left'] . $params['title'] . $params['like_right'] .'"';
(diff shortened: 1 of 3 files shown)


notartom:
Heya,

Thanks for the PR!

It is indeed currently impossible to set up a "dev" LibriVox on one's machine - while the Ansible playbooks would work, they depend on things like database dumps that obviously only exist in staging/production. Enabling the playbooks to run on anyone's dev machine is next in my LV plans.

Let me find some time to properly review this and test it in our dev environment (we do have one, but I obviously can't give anyone and everyone access), and then hopefully we can merge this quickly.

Cheers!

rillig:
I updated the commit. The previous commit was not useful since it changed the CodeIgniter code, which would make updating needlessly complicated.

notartom:
Thanks for tackling this! Looks like searching by keywords in advanced search is also broken, if you're looking for more things to do ;)

PR #5, opened 2018-03-27
Fix advanced search for section language

rillig:
Given the following scenario:

    - a multilingual project
      - an Arabic section
      - a Chinese section

When I browse for "language Arabic" in the catalog, the URL is:

    /search/get_results
      ?primary_key=4
      &search_category=language
      &sub_category=language

and the search result is:

    - a multilingual project
    - an Arabic section

But when I do an "Advanced search" for "language Arabic", the URL is:

    /advanced_search
      ?recorded_language=4
      &search_form=advanced

and the search result is:

    - a multilingual project
    - an Arabic section
    - a Chinese section

This is unexpected.

This commit makes the two search results equal by only showing sections
of the selected language. Sections in other languages that happen to be
part of a multilingual project containing the selected language are no
longer shown.

### DIFF
--- a/application/libraries/Librivox_search.php
+++ b/application/libraries/Librivox_search.php
@@ -73,14 +73,20 @@ class Librivox_search{
 
 		$language = ' ';
 		$language_clause = ' ';
-		if (!empty($params['recorded_language']))
+		$section_language_clause = ' ';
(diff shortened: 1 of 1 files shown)


notartom:
Will be handled on GerritHub, closing.

https://review.gerrithub.io/#/c/405997/

PR #9, opened 2019-07-31
Improve RSS sorting

Cj-Malone:
Currently some podcast clients have issues displaying LibriVox recordings in the correct order, this is presumably because you don't have `pubDate` like standard podcasts.

However LibriVox can provide `section_number` as the episode number so podcast clients can use that to sort.

I have also added iTunes `type` = `serial`, this is used to identify podcasts where order is important, apposed to the majority of podcasts where they are time relevant, and the latest episode is emphasised. This should help compatible podcast clients prompt the user to listen from the start, in the correct order.

NOTE: It looks fine to me but I don't have a PHP environment set up so I can't test this. The only potential issue I can see would be `section_number` as I don't fully understand the objects in this project. But I'm fairly confidant this should work fine.

### DIFF
--- a/application/views/rss/rss.php
+++ b/application/views/rss/rss.php
@@ -8,6 +8,7 @@
   <description><![CDATA[<?= strip_tags($project->description) ?>]]></description>
   <!--<genre>project element=Genre</genre>-->
   <!--<language>project element=lang.code</language>-->
+  <itunes:type>serial</itunes:type>
   <itunes:author>LibriVox</itunes:author>
   <itunes:summary><![CDATA[<?= strip_tags($project->description) ?>]]></itunes:summary>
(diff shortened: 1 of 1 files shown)


notartom:
Heya, thanks for the PR! Could I ask you to propose it as a Gerrit review, as per [1]? Cheers!

[1] https://github.com/LibriVox/librivox-catalog/blob/master/CONTRIBUTING.md

Cj-Malone:
Hey, [this](https://review.gerrithub.io/c/LibriVox/librivox-catalog/+/463932) Has been sitting in GerritHub for almost a month now, is there anything more I can do?

notartom:
Sorry for taking a month to get to this. It's now merged on gerrithub.

PR #10, opened 2020-06-13
Include cover art on podcast feeds.

colons:


kgroeneveld (commented):
I agree it would nice to include the cover art in the RSS feed.

Some comments:

The `coverart_jpg` field is empty for some items in the catalog. Most other spots where the cover art is displayed will substitute a generic image in this case. I wonder if the RSS feed should do the same or if the RSS feed should just omit the image tag in this case? The pull request as is will result in an empty href in this case which doesn’t seem good.

The Apple specs for the `<itunes:image>` tag (https://help.apple.com/itc/podcasts_connect/#/itcb54353390) specify the size must be minimum 1400x1400. I think most of the LibriVox images are a lot smaller than this. I am not sure how much that would matter.

I wonder if it would be better to use a standard RSS `<image>` instead of the itunes one? Or maybe include both? I don’t really have much experience with RSS...

notartom:
Thanks for the PR, and sorry for the delay in reviewing it. This was slightly improved by @hornc and done in https://github.com/LibriVox/librivox-catalog/pull/163.

PR #11, opened 2020-06-21
Refactor search code

notartom:
Replace 5 instances of almost-the-same code with a simple function.


kgroeneveld (approved):
I am a fan of putting repeated code in a function. I tested this change and it all seems good. The new function also has the added advantage of not including an extra zero element at the beginning of the id arrays when it is not needed.

While testing and reviewing this I noticed that there is an extra unused call to _get_projects_by_section_author at line 149. Looks like the left overs of a copy/paste/edit. Maybe you can remove this as part of this refactor?

redrun45:
I did a merge to the current master, and this refactor has identical functionality.  I assume the extra call was removed elsewhere in the meantime.  Here's my :+1:.

PR #12, opened 2020-06-21
Fix advanced language search for compilations

notartom:
There are compilations where every section has the same language as the
corresponding project (examples: Sammlung deutscher Gedichte). When
searching by language, the individual sections are not interesting,
only the project itself should be listed.

### DIFF
--- a/application/libraries/Librivox_search.php
+++ b/application/libraries/Librivox_search.php
@@ -85,8 +85,8 @@ class Librivox_search{
 
 			$language_clause = ' AND ( p.language_id = ' . $recorded_language . ' OR p.id IN (' . implode(',', $section_project_ids) . ') ) ';
 			$section_language_clause = '
-				AND ( p.language_id = ' . $recorded_language . '
-					OR st.section_language_id = ' . $recorded_language . ') ';
+				AND ( p.language_id <> ' . $recorded_language . '
+					AND st.section_language_id = ' . $recorded_language . ') ';
 		}
 
 


notartom:
Salvaged this from GerritHub, originally proposed by @rillig

redrun45:
A "what does this do" review, the best I know how:
It took me a while to spot the change in behavior, but that might be due to more recent changes in master.  I've tested using just this patch on an otherwise fresh master.

When doing an _Advanced Search_, filtering by Language -> German, and NOT searching by title or author, I see the difference.  The individual sections of these "Sammlung deutscher Gedichte" collection projects do not show up in the results.  Instead, since they are part of _collection_ projects and those projects have been added to a _group_, only that top-level "Sammlung deutscher Gedichte" _group_ shows up in the results.

I went looking for potential problems, and I found the scope of this is pretty narrow:
- When doing a search (advanced or otherwise, language-filtered or otherwise) for the title or author of one of these sections, the individual section does show up as a result, as expected.
- When doing a search (again, filtered or not) for the title of one of these _projects_, the _group_ that all the projects have been collected under shows up instead, as expected.
- When merely _browsing_ by language (clicking the little 'Language' flag in the top, then selecting 'German', for example), the results are unchanged.  The sections of the Sammlung deutscher Gedichte projects show up individually in alphanumeric order, as in current master.  I don't know whether this is expected, or if they ought to match the change in _Advanced Search_ behavior.

redrun45:
@rillig, since I see you're around, does my description match the intended behavior?

If so, just let me know, and I'll let the multi-lingual admins decide if they find this useful.  😉

rillig:
It's been a while since I wrote this small patch. What you found out is what I intended, so yes, please suggest to apply it.

redrun45:
Admins do like this idea for tidying up search results re: collections, so I've gone back for another, closer, look.

On closer inspection, this does seem to remove some desired results.  For example, searching for a section in an all-German collection, by language and by title:
https://librivox.org/search?title=Rede+zur+Er%C3%B6ffnung&author=&reader=&keywords=&genre_id=0&status=all&project_type=either&recorded_language=3&sort_order=alpha&search_page=1&search_form=advanced

On live, I get one result; on this branch, none.  The intended behavior is to show the section regardless of the collection project's language.

application/libraries/Librivox_search.php
```diff
@@ -85,8 +85,8 @@ function advanced_title_search($params)
 
 			$language_clause = ' AND ( p.language_id = ' . $recorded_language . ' OR p.id IN (' . implode(',', $section_project_ids) . ') ) ';
 			$section_language_clause = '
-				AND ( p.language_id = ' . $recorded_language . '
-					OR st.section_language_id = ' . $recorded_language . ') ';
+				AND ( p.language_id <> ' . $recorded_language . '
+					AND st.section_language_id = ' . $recorded_language . ') ';
```
redrun45, line 89:
```suggestion
				AND st.section_language_id = ' . $recorded_language;
```
If I'm holding all of the pieces right in my head, I think we can do away with this first clause.  It would seem to discard all sections of a collection project, where the entire collection is in the language we actually want.

I'm new to the GitHub code review process, particularly from this end.  I can always open a new PR, or I can test this one again if you'd like to change it.

notartom: So apparently I can commit this suggestion (still new to the less frequently used functions of the PR workflow myself), but I'd rather not have two commits for a really simple change. Which I guess means you get to open a new PR (crediting @rillig of course).

redrun45: Sounds good!  I'll file a new PR, most likely tomorrow evening.  :+1:

redrun45: Meh, I have not been holding the pieces correctly in my head.  My suggested change leaves both the Collection projects, _and_ all of their constituent Sections, in the results.  I'm in this deep, so I'll likely still come back with a PR soon, it just won't be this one.

I have a lot of respect for folks who can do this for a living.

Issue #7, opened 2019-04-06
Upgrade to CodeIgniter 3


notartom:
We're currently on CodeIgniter 2.1.1. In addition to the obvious, this prevents us from moving away from the EOL'ed PHP5.6 and upgrading to PHP7.

The upgrade guide is here: https://www.codeigniter.com/userguide3/installation/upgrade_300.html

kgroeneveld:
There may very well be issues using CodeIgniter 2.1.1 with newer versions of PHP but so far on my dev setup I have not noticed an issue while using PHP 7.4.3.

notartom:
Huh, I wonder where I got the CodeIgniter 3 == PHP 7 requirement from. Even their own [doc](https://codeigniter.com/userguide3/general/requirements.html) says it should work with PHP5.6.

Out of curiosity, how did you set up your dev environment?

kgroeneveld:
I setup my dev environment under kubuntu 20.04. I used the librivox-ansible scripts and hacked and slashed them until it worked. Such as changing all the `php5.6` references to just `php` among other various things I probably should have made better notes on.

I think the bigger issue is does the old CodeIgnitor work properly with PHP 7, not whether CodeIgnitor 3 needs PHP 7. It would be nice to update the base operating system from ubuntu 16.04 to the latest LTS. Ubuntu 20.04 no longer has php5.6 (although there is probably some PPA for it or you could probably build it yourself).

Of course it would also be good to update CodeIgnitor eventually. CodeIgnitor 4 is out now so this issue should maybe be updated to be “Upgrade to CodeIgniter 4”.

It seems you have already done a bunch of work relating to the official CodeIgnitor 3 upgrade guide. I did want to eventually review/test that pull request in more detail but I am not sure when / if I will find the time. I am a LibriVox, CodeIgnitor and PHP newbie so it takes a lot of effort to review it properly.

There is likely some issue somewhere with using PHP 7 with the current LibriVox catalog. I have really only been testing search/browse and project launch templates so far. It could be horribly broken elsewhere. I didn’t intend to imply in my previous post that I think everything is OK just that it is not completely broken either.

I did have some issues relating to the newer mariadb in ubnuntu 20.04. I have been thinking of creating a new issue for that.

And it would be nice to update jquery, and basecamp, and, and …  It’s a lot of work to keep everything up to date. And I do often think if it ain’t broke don’t fix it.

notartom:
[...]

PR #13, opened 2020-06-21
CodeIgniter 3 upgrade

notartom:
Issue #7


application/libraries/MY_Form_validation.php
```diff
@@ -1,10 +1,11 @@
 <?php
 
-class MY_Form_Validation extends CI_Form_Validation
+class MY_Form_validation extends CI_Form_validation
 {
 	function alpha_dash_space($str)
     {
+        echo("<!-- alpha_dash_space $str --!>");
```
kgroeneveld, line 7: Is this just left over debug code that was accidentally added?

notartom: Yep, good spot.

kgroeneveld:
As I mentioned in #13 it seems that CI3 expects there to be an `error_exception.php` file that we do not have. I also just noticed that there is a reference implementation of this in the CI3 release and there are also updated reference error views. It does not really seem that our code base customized these so I would suggest we copy all the references files from `application/views/errors/html` to our project.

kgroeneveld:
There are actually a lot of files in the reference application directory that we should maybe consider if we should also update. For example things like `config/foreign_chars.php` or other files in the `config` directory. But we can't just copy them all blindly as some of them such as `config/constants.php` have been customized in our project. I am not sure if it is worth the bother to investigate all of these.

kgroeneveld (commented):
I don't really know enough about CodeIgnitor to declare this a non mystery but I think class names starting with `CI_` are meant to be used for classes that are part of CodeIgnitor itself. If you look at the CI2 source after loading a file it tries to look for a class starting with `CI_` or just the base class name that was requested. The CI3 source only tries the base class name. I think the `Template` class was just always named wrong and CI2 was more forgiving.

application/views/catalog/partials/footer.php
```diff
@@ -19,18 +19,18 @@
...
-	var search_page = <?= (int)($search_page ?: 1) ?>;
+	var search_page = <?= (int)(isset($search_page) ?: 1) ?>;
 	set_advanced_form_page(search_page);
 
 	var search_order = 'alpha';
 
 	var project_type = 'either';
 
-	var q = <?= json_encode($search_value ?: '', JSON_HEX_TAG | JSON_HEX_QUOT) ?>;
+	var q = <?= json_encode(isset($search_value) ?: '', JSON_HEX_TAG | JSON_HEX_QUOT) ?>;
```
kgroeneveld, line 33: I think this should be:
`var q = <?= json_encode(isset($search_value) ? $search_value : '', JSON_HEX_TAG | JSON_HEX_QUOT) ?>;`

Making that change seems to fix the search from the home page always searching for "true" (see my recent comments in #106).

I expect the other two changes in this commit should be updated similarly.

I do wonder if this is even the best way to fix this at all or if code somewhere else should be changed to always set these variables. I don't understand the overflow well enough to say one way or the other right now.

notartom: > I think this should be:
> `var q = <?= json_encode(isset($search_value) ? $search_value : '', JSON_HEX_TAG | JSON_HEX_QUOT) ?>;`


Yep, good catch. As it is, using the shorthand ```?:``` means the expression evaluates to the value of the first operand if the latter is True-ish, and because we've started using ```isset``` that's just ```True```

PR #14, opened 2020-06-21
Display author information concisely and update it correctly

notartom:
If an author has 0 completed books, that information is not worth being
displayed. Same for 0 books in progress. Use correct plural forms for
the above information (book/books).

Authors that had "Completed: 0 books | In progress: 0 books" before do
not have any information at all now. This can later be extended to count
the sections instead of the projects, for those authors that only appear
in collective works.

When updating the book statistics, also update those authors that don't
have any books at all. Without this extra cleanup, these authors will
keep the previous count of their books, which is wrong. To see the wrong
behavior:

    UPDATE authors SET meta_complete = -1, meta_in_progress = -1;

    https://librivox.org/cron/project_status_stats/author

    https://librivox.org/search

Now some of the authors still have the -1 count.

### DIFF
--- a/application/controllers/cron/project_status_stats.php
+++ b/application/controllers/cron/project_status_stats.php
@@ -22,6 +22,11 @@ class Project_status_stats extends CI_Controller
 
 	public function author()
 	{
+		$this->db->query('
+			UPDATE authors
+			SET meta_complete = 0,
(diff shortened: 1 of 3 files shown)


notartom:
Replaced by #242

PR #18, opened 2020-07-17
add issue templates

RayBB:
Adding issue templates.

I think these are at least a small step to make it a little easier for someone to contribute in terms of reporting a bug. Templates are pretty common in other popular repos. We discussed it a bit [here](https://forum.librivox.org/viewtopic.php?f=17&t=81235).

These will show up when someone clicks the "new issue" button.

No changes are needed in the repo setting. It will function like the example in my repo [here](https://github.com/RayBB/issue-template-test).

Let me know if there's a better way to phrase the template, this is just a starting point.

### DIFF
--- a/ISSUE_TEMPLATE/feature_time.md
+++ b/ISSUE_TEMPLATE/feature_time.md
@@ -1,0 +1,43 @@
+---
+name: Feature time
+about: Create a report to help us decide what to do next
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+
+### Is your proposal related to a problem?
+
+<!--
+  Provide a clear and concise description of what the problem is.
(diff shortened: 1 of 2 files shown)


ISSUE_TEMPLATE/feature_time.md
```diff
@@ -0,0 +1,43 @@
+---
+name: Feature time
```
notartom, line 2: I'm nitpicking here, but why "time"? I can't help up think of Adventure Time when I read this. Why not Feature request?

RayBB: So sorry about not responding sooner. Somehow the notification for this slipped though the cracks for me. I updated this in a new PR.

PR #21, opened 2021-01-03
Fix image link href in browse catalog title blade.

kgroeneveld:
The second commit does not apply without the first so I made them one pull request even though they are separate (but related) issues.

### DIFF
--- a/application/views/catalog/item_blades/title_blade.php
+++ b/application/views/catalog/item_blades/title_blade.php
@@ -4,11 +4,24 @@
 		<div class="catalog-type"><span class="title-icon"></span>Title</div>
 	<?php endif; ?>
 
+	<?php
+		if ($item['status'] == PROJECT_STATUS_COMPLETE)
+		{
+			if (empty($item['coverart_thumbnail']))
+				$image = base_url() . 'images/book-cover-default-65x65.gif';
+			else
+				$image = $item['coverart_thumbnail'];
 
-	<?php $image = (empty($item['coverart_thumbnail']))? base_url().'images/book-cover-default-65x65.gif': $item['coverart_thumbnail'] ;?>
-	<?php $image = ($item['status'] == 'complete') ? $image  : base_url().'images/book-cover-in-progress-65x65.gif' ?>			
+			$url = $item['url_librivox'];
+		}
+		else
+		{
+			$image = base_url() . 'images/book-cover-in-progress-65x65.gif';
+			$url = $item['url_forum'];
+		}
+	?>
 
-	<a href="#" class="book-cover"><img src="<?= $image ?>" alt="book-cover-65x65" width="65" height="65" /></a>	
+	<a href="<?= $url ?>" class="book-cover"><img src="<?= $image ?>" alt="book-cover-65x65" width="65" height="65" /></a>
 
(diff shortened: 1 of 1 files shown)


notartom (changes requested):
I don't think this commit is necessary at all, looks like all our coverart_thumbnail URLs are for archive.org, and can be updated to HTTP directly in the database:

`artom@librivox:~$ mysql librivox_catalog -e 'select coverart_thumbnail from projects where coverart_thumbnail<>""'  | grep -v archive\\\.org
coverart_thumbnail
artom@librivox:~$`

kgroeneveld:
I dropped the second commit from the PR. (Hopefully I did it correctly, I have never edited a github pull request before.) It probably should have been a separate request anyway.

I agree that it would be good to update all the entries in the data base to be https.  (I even wonder if the data base should include the https:// at all or if that should be inserted at runtime, but that is a bigger change and may not even make sense.) My only concern was that we should then be preventing new entries from having same issue then. My patch was really only meant to be a temporary solution.

notartom:
So where do these http:// links appear in the HTML? They're definitely http:// in the database, but something appears to be translating them to https:// before they hit the view, at least in all the places that I'm looking.

kgroeneveld:
Try going to https://librivox.org/the-30000-bequest-and-other-stories-by-mark-twain/.  Firefox warns about mixed content. If you view the page source you can see the title artwork is `book-cover-large`.  (Assuming you are not running the original version of this PR which converted it to https.)

notartom:
OK - next question is - where is the http:// coming from when it's saved in the database? Because as you've said, doing s/http/https/ in the DB is one thing, but it won't prevent new http:// URLs from being added. I've done a few greps around the codebase, and I can't figure it out. I suspect `coverart_thumbnail` is not being saved explicitly, but rather as an ORM save call, I just don't know where that happens.

kgroeneveld:
>OK - next question is - where is the http:// coming from when it's saved in the database?

Good question. I am not sure exactly myself. I haven't quite figured out the typical flow in CI / LibriVox catalog yet. I think it is input by the user in this view: `application/views/private/administer_projects/add_catalog_item.php`. And I think that form is validated in `application/controllers/private/administer_projects.php`. Maybe it would be a matter of adding another validator similar to the one for Title in that file: `$this->form_validation->set_rules('projectname', 'Title', 'trim|required|xss_clean');`.

There is a reason I started off by just hacking the code to dynamically change the http to https...  I was looking for a quick fix to the security warnings browsers are giving now and I didn't really know how to fix it properly yet. I also was not sure how easy it would be to fix the entries in the existing database. That isn't something I have access to do even if I was confident about how to do it. I guess it could be done with a fairly simple script one could test on a copy of the database.

I wonder how much of the https discussion should be moved to #26 and start a separate PR once we decide on a plan. The other remaining commit in this PR is really a separate issue. In hindsight I regret lumping them together.

notartom:
> I wonder how much of the https discussion should be moved to #26 and start a separate PR once we decide on a plan. The other remaining commit in this PR is really a separate issue. In hindsight I regret lumping them together.

Good point. Let's get this one merged, and carry on the HTTP(S) discussion in #26.

PR #22, opened 2021-01-03
Fix searches that contain certain special symbols.

kgroeneveld:
There were two spots causing issues:
1. The ajax get request to perform the actual search was not URI encoding
   the query. This still worked for most symbols but some such as "&" and
   "+" were failing.
2. When performing a search from a page other than the search results
   page it was encoding the search query into the URL path in the form of
   "search/q/{query}". But this fails for many different symbols. Even if
   you try to URI encode the query a "/" will still break it. This commit
   changes the URL to be "search?search_form=get_results&q={query}" and
   URI encodes the query. This seems to work fine for every possible
   query with special symbols I could think of. This exact same issue
   is also present in the Wordpress theme.

### DIFF
--- a/application/views/catalog/partials/footer.php
+++ b/application/views/catalog/partials/footer.php
@@ -267,7 +267,7 @@
 		$.ajax({
 		  	url: CI_ROOT  + 'advanced_search',
 		  	type: 'get',
-		  	data: $('#advanced_search_form').serialize() + '&q=' + q ,
+		  	data: $('#advanced_search_form').serialize() + '&q=' + encodeURIComponent(q),
 		  	beforeSend: function(){
 		  		$('.browse-list').html(spinner);
 		  		$('.page-number-nav').html('');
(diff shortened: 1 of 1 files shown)


notartom:
Playing around with this, I wasn't able to reproduce the problem that this is supposed to fix. AFAICT, the only place where searches happen is either the "simple" search box in the top right in the header, or the advanced search that expands out when you click the link below the "simple" search box. In both of these, characters like & and + appear to work fine.

kgroeneveld:
I just tried doing a search from the simple search box for `&` and it seems to return every item in the catalog, not just ones that contain `&`.

Go to (for example) https://librivox.org/a-dogs-tale-by-mark-twain/. Search fro `dog's`.  It gives an error. You will also get the same error if you do the search from the LibriVox home page.  If you do that same search from the Browse The Catalog / search results page it works okay.

notartom:
It's a partial fix in that it only fixes searches from book pages, and not the home page or search results, but it is an improvement, so LGTM.

kgroeneveld:
A bit of a side note… My personal preference is that PR’s for short lived topic branches are generally done with a rebase and fast forward commit instead of a merge. I think it makes the git history far neater.

notartom:
> A bit of a side note… My personal preference is that PR’s for short lived topic branches are generally done with a rebase and fast forward commit instead of a merge. I think it makes the git history far neater.

I'd agree - I'm still figuring out the GitHub review system, your comment made me look and learn the 3 different ways of accepting PRs.

Issue #41, opened 2021-01-13
Change the status to trigger generation of the catalog URL on the project screen


twinkietoes-on:
Currently the Catalog URL is generated when marking the project Complete. However, this URL field needs to be populated before the files are sent to Archive because the description there includes this URL. Currently the MCs are marking the project complete, then immediately marking it as in Validation - IF they remember to do so. Otherwise they have to remember to get the URL after the project is marked Complete and modify the description at Archive to add it in.

Priority: 2

PR #75, opened 2021-01-19
Generate catalog URL when status changes to Validation.

kgroeneveld:
Previously the URL was generated when the status changed to Complete.

This implements the request from issue #41.


notartom:
For the record, since I'm not at all involved in the actual audiobook-making process of LibriVox, I trust you implicitly on this.

PR #76, opened 2021-01-25
Open links in Magic Window in a new tab/window.

kgroeneveld:
Opening these links inside the iframe definitely doesn't seem ideal.
Opening them in the same window (outside the iframe) might make sense.
But I think most people would prefer a new tab/window.

### DIFF
--- a/application/views/public/public_readers_iframe/index.php
+++ b/application/views/public/public_readers_iframe/index.php
@@ -2,7 +2,7 @@
 <head>
 	<title>Librivox Project iFrame</title>
 	<meta charset="utf-8">
-
+	<base target="_blank">
 
 	<style type="text/css">
 


twinkietoes-on:
FWIW, 3 of 3 admins who responded (plus me) are in favor of this: https://forum.librivox.org/viewtopic.php?p=1842380#p1842380

notartom:
Fair enough, you're the ones working with this :)

FWIW, I'm against breaking web convention in this way. There's already a way to open links in new windows/tabs (with middle click)  that gives the user the choice of same window/tab or new one.

kgroeneveld:
>FWIW, I'm against breaking web convention in this way. There's already a way to open links in new windows/tabs (with middle click) that gives the user the choice of same window/tab or new one.

In general I agree with that. But in this case when it is in an iframe it just seems weird to open a full web page inside the relatively small magic window.

notartom:
> > FWIW, I'm against breaking web convention in this way. There's already a way to open links in new windows/tabs (with middle click) that gives the user the choice of same window/tab or new one.
> 
> In general I agree with that. But in this case when it is in an iframe it just seems weird to open a full web page inside the relatively small magic window.

Ah, I wasn't aware of the iframe thing. In that case, yeah, new tab/windows sounds better.

Issue #27, opened 2021-01-10
Add Check for _128kb at end of file name to "run tests" in Validator


twinkietoes-on:
File names are supposed to end in "_128kb.mp3". A validator test to check for this would be helpful.

Priority: 1

kgroeneveld:
I have started to look into this one. I think I found the spot to change but I still need to dig into it further to make sure.

PR #78, opened 2021-01-30
Validator improvements including new check for "_128kb" in filename.

kgroeneveld:
The first commit in this PR addresses #27.

While working on that I found a number of other issues in the validator that were bugging me so this PR includes a couple more commits for some of those issues. It seemed as though I found an endless number of issues and confusingly written code in the validator such that I could spend much more time on it. I tried to keep the changes to a minimum of low risk changes and resist my urge to re-write larger sections of it (and make my own mess)...  maybe another time.

See commit logs for more details on the actual changes.

### DIFF
--- a/application/libraries/Librivox_filecheck.php
+++ b/application/libraries/Librivox_filecheck.php
@@ -51,7 +51,8 @@ class Librivox_filecheck{
 
 	public function load_project_map($validation_dir, $map, $project)
 	{
-		if (empty($validation_dir) || empty($map)) return false;
+		// an empty $map is ok and expected when there are no files (which will be flagged in the completeness checks)
+		if (empty($validation_dir) || !is_object($project)) return false;
 
 		$this->validation_dir = $validation_dir;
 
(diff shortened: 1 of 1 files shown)


kgroeneveld:
After I switched branches in my local git repo I seem be getting an error relating to this PR. I should probably double check what is going on before this gets merged...

kgroeneveld:
>After I switched branches in my local git repo I seem be getting an error relating to this PR. I should probably double check what is going on before this gets merged...

False alarm. When I switched branches I lost an un-commited changed that breaks the validator on PHP 7.4. If I re-instate that change or checkout this exact PR branch on my other dev system with PHP 5.6 it works fine.

twinkietoes-on:
For the record, I have tested this, and it seems to work without breaking anything else. ;)

notartom:
I'm assuming the bitrate has to be 128kpbs, and that has to be reflected in the filename?

twinkietoes-on:
Yes. Archive needs that appended onto the file names in order to identify the MP3s as 128 kbps. Otherwise they're identified as VBR.

PR #88, opened 2021-02-12
Add Simplified Chinese and update Portuguese lanugages.

kgroeneveld:


notartom:
Let's hope our stack's unicode support is top notch ;)

Issue #55, opened 2021-01-16
Remove (-/-) when no data entered for birth/death dates in author table


twinkietoes-on:
In catalog lists and book pages

Examples: 
https://librivox.org/campfire-tales-of-jackson-hole-by-g-bryan-harry/
https://librivox.org/author/15350 (it's not showing under the author name, but still showing on the book listing)

Priority: 3

RayBB:
[This](https://github.com/LibriVox/librivox-catalog/blob/58e96e31873435145c392a60eb05b964190c7114/application/helpers/previewer_helper.php#L198) seems to be the function in question.
It is only referenced three places in the codebase:
Author page - [under author's name](https://github.com/LibriVox/librivox-catalog/blob/58e96e31873435145c392a60eb05b964190c7114/application/views/catalog/author.php#L13)
Author page - [book listings](https://github.com/LibriVox/librivox-catalog/blob/58e96e31873435145c392a60eb05b964190c7114/application/views/catalog/item_blades/author_blade.php#L8)
Book page - [next to author's name](https://github.com/LibriVox/librivox-catalog/blob/58e96e31873435145c392a60eb05b964190c7114/application/views/catalog/item_blades/title_blade.php#L33)

It's not really clear to me how what might be causing the issue here since the usage is pretty similar.
The main difference seems to be that the two that are having issues are both [blades](https://laravel-guide.readthedocs.io/en/latest/blade/).

I'm not familiar enough with php or laravel to understand what is causing the difference when using blades.
My best guess is that the blades are being passed different data somehow.

kgroeneveld:
I had started to look at this issue as well. I found at least four different functions that format the birth/death dates (and that isn't even including the one you linked to). I was planning on trying to unify them all as much as possible and have various changes sitting locally on my computer. Not sure when I will have time to finish it. I am easily distracting working on multiple things at once. I may look at it again this weekend.

kgroeneveld:
I am really starting to go down the rabbit whole on this one...  

I have found 12 different spots that format the author dates now with many different variations on how it is implemented.

There are 4 examples in the JS code.  Two of these seemed to do the correct thing, one did not (it would show `(null - null)`) and one did not include the years at all when it should have.  I have consolidated all the JS uses down to one function.  While working on that I also found a couple bugs with author/translator suggestions on the new project page which got me a bit side tracked.

[...]

PR #93, opened 2021-02-13
Author formatting

kgroeneveld:
The original motivation for this change was issue #55 but I ended up going way beyond and refactoring all of the author display formatting code to fix many other inconsistencies and make some other formatting changes after consulting with the admins.

See the commit logs for many more details.

I have a related change for the WordPress code (that actually depends on this change) that I will submit once this gets merged. Technically issue #55 will not be fully addressed until the WordPress change is in place as well.

### DIFF
--- a/application/helpers/previewer_helper.php
+++ b/application/helpers/previewer_helper.php
@@ -10,12 +10,6 @@ function format_filename($string)
 	return $transliterator->transliterate($string);
 }
 
-function author_name($fields)
-{
-	$fields['dob'] = (empty($fields['dob'])) ? '' : $fields['dob'];
-	return $fields['first_name']. " " .$fields['last_name']. " (" . $fields['dob'] . " - " . $fields['dod'] . ")";
-}
-
 function clean_title($title, $retval = 'title')
 {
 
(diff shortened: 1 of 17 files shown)


application/helpers/previewer_helper.php
```diff
@@ -139,39 +130,100 @@ function create_title_slug($project)
...
-function get_language_code($language)
+define ('FMT_AUTH_YEARS', 0x01);
+define ('FMT_AUTH_HTML',  0x02);
+define ('FMT_AUTH_LINK',  0x04);
+define ('FMT_AUTH_WIKI',  0x08);
+
+function format_authors($authors, $flags = 0, $max = 0)
 {
-	if (empty($language)) return '';
+	$flag_years = (bool)($flags & FMT_AUTH_YEARS);
```
notartom, line 167: This is such a C thing to do ;)

kgroeneveld: >This is such a C thing to do ;)

Is that a bad thing? Or a good thing?

I actually got the idea partially from PHP functions like https://www.php.net/manual/en/function.mb-convert-case.php which use defined constants to pass in different options. But I just looked again and I see you can only pass in _one_ constant to that function. 

I wonder if any standard PHP functions use an ORed options mask like I used? 

Is there a commonly accepted way to do something similar in PHP?

Languages like Perl often use a hash to pass in various options like this. I guess you could use an (associative) array to do something similar in PHP. But when you learned to program on a computer with 48kB of RAM these methods just don't sit well with me...  they cause global warming.

notartom: > > This is such a C thing to do ;)
> 
> Is that a bad thing? Or a good thing?

It's just... a thing :) It shows your origins, and experience, but is neither good nor bad. 

> I actually got the idea partially from PHP functions like https://www.php.net/manual/en/function.mb-convert-case.php which use defined constants to pass in different options. But I just looked again and I see you can only pass in _one_ constant to that function.
> 
> I wonder if any standard PHP functions use an ORed options mask like I used?
> 
> Is there a commonly accepted way to do something similar in PHP?

I don't know PHP well enough to have an informed opinion. I think in Python we'd just have a bunch of boolean keyword arguments with default values, something like:

```python
def format_authors(authors, years=False, html=False, link=False, wiki=False):
```

I believe something similar is possible with PHP.

> Languages like Perl often use a hash to pass in various options like this. I guess you could use an (associative) array to do something similar in PHP. But when you learned to program on a computer with 48kB of RAM these methods just don't sit well with me... they cause global warming.

I... don't have an answer to that. Or maybe I do, but it's very long and philosophical, and involves how my kids grow up in a world where "computers" are for solely consuming media rather than tinkering and creating or programming, and is more appropriate for a pub than a pull request.

kgroeneveld: >def format_authors(authors, years=False, html=False, link=False, wiki=False):

That is possible in PHP as well. But the annoying thing with that is you need to remember the order of the parameters when using the function add add a bunch of extra false parameter values to use the last option.

Anyway... that is probably enough somewhat off topic discussion in this PR.

PR #96, opened 2021-02-14
Fix recent glitch in display of birth/death years.

kgroeneveld:
It was only a matter of time before I broke something... fortunately this is fairly minor.

And now that the WordPress theme uses this same code directly I only need to fix it in one spot. The JavaScript version of this function was okay.

### DIFF
--- a/application/helpers/previewer_helper.php
+++ b/application/helpers/previewer_helper.php
@@ -151,7 +151,7 @@ function format_author_years($author)
 	$dob = empty($author->dob) ? '' : $author->dob;
 	$dod = empty($author->dod) ? '' : $author->dod;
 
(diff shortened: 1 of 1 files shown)


notartom:
Huh, I actually might have caught this in review, had I looked closer.

kgroeneveld:
>Huh, I actually might have caught this in review, had I looked closer.

It takes a lot of effort to see a bug like that in a commit that large. Especially when dod vs dob used multiple times on the same line of code look so similar.

And I did a lot of testing of the code as well as a couple admins tested it. It wasn't until the WordPress change went live and @twinkietoes-on tested that and noticed one of the authors did not have the years portion display when it should have.

PR #105, opened 2021-02-28
Add Russian translation

kgroeneveld:
This also includes a small cleanup with colon usage on the uploader page.


application/language/russian/project_launch_lang.php
```diff
@@ -0,0 +1,127 @@
...
+$lang['proj_launch_additional_info_project'] = 'Дополнительные сведения о проекте';
+$lang['proj_launch_additional_info_soloist'] = 'Сведения о вас (если это ваш первый сольный проект)';
+$lang['proj_launch_choose_lang'] = 'Выберите язык для вашего шаблона';
+$lang['proj_launch_project_solo'] = 'Сольный проект';
+$lang['proj_launch_project_collaborative'] = 'Совместный проект';
+$lang['proj_launch_project_dramatic'] = 'Чтение по ролям';
+$lang['proj_launch_project_poetry_weekly'] = 'Еженедельное стихотворение';
+$lang['proj_launch_project_poetry_fortnightly'] = 'Двухнедельное стихотворение';
+$lang['proj_launch_fill_form'] = 'Заполните эту форму как можно полнее. Скопируйте и вставьте в новый пост форума New Projects Launch Pad.';
+$lang['proj_launch_type_of_project'] = '<span class="red">*</span> Тип проекта';
```
notartom, line 22: I wonder if we could move the HTML out of the translations?

notartom (changes requested):
Hey, something I'm actually qualified to review :)

I realize the actual Russian bits aren't from you, so I'll take the liberty of making the small fixes myself to make things move quicker.

notartom:
I've addressed my comments in https://github.com/LibriVox/librivox-catalog/commit/17c94395c8574dec9d3a5b522ffedee655580e45.

application/language/russian/project_launch_lang.php
kgroeneveld: I agree that all the `<span class="red">*</span>` HTML should be in the PHP code and not in the translations. I was close to making this change before when doing other work on the Project Launch and language stuff. Maybe now that you have expressed a similar opinion I will go ahead and make this change at some point.

Issue #109, opened 2021-04-05
Errors in SOME compilation project catalog pages


twinkietoes-on:
This project has errors in the author field:
https://librivox.org/worldstory-volume-xiv-outline-of-universal-history-by-karl-ploetz/

But the other compilations I'm looking at do not:
https://librivox.org/coffee-break-collection-30-mythical-creatures-by-various/

The one with the error (1) has over 100 sections; (2) does not have cover art yet. It's a new project, cataloged yesterday. Those are the only differences I've come up with, and it would be strange to have them be the cause!

twinkietoes-on:
Aha! It seems to only occur on compilations where the author is unassigned. I cannot think of a project where this occurs except in error. (This specific project, I think, shouldn't be a compilation.)

notartom:
So the fix would be to set the correct project type on those authorless compilations (that aren't actually complication?)

twinkietoes-on:
I don't know. I have to think this through a bit more. I think in this case it's user error (no authors + no text in the sections =  not a compilation). But I am trying to remember if there are cases where it's a compilation with separate texts but no author in the sections. Example would be a single author project with multiple texts, which may not have the author put in the sections.

Definitely no author + no text = not a compilation, but not sure about just no author.

notartom:
Well, I've haxxed some `isset`s on the live server (I know, not great, but our staging database snapshot is out of date and doesn't have that project to test with) to at least just show the empty string (and not the PHP error). This is obviously temporary.

kgroeneveld:
>Well, I've haxxed some issets on the live server (I know, not great, but our staging database snapshot is out of date and doesn't have that project to test with) to at least just show the empty string (and not the PHP error). This is obviously temporary.

I submitted a pull request before your post that adds an `isset`. I tested it by creating a compilation on my test server with a section with no author.  Whether this should be considered a valid project and allowed to be created is another question, but adding a check for null author in `format_author` seems reasonable in any case.

notartom:
Fixed by PR#110

PR #110, opened 2021-04-05
Check for null author passed into format_author().

kgroeneveld:
One such case where null gets passed to this function is when viewing
a compilation where a section has no author set. This would result in
errors such as:

A PHP Error was encountered
Severity: Notice
Message: Trying to get property of non-object
Filename: helpers/previewer_helper.php
Line Number: 149

And:
A PHP Error was encountered
Severity: Notice
Message: Trying to get property of non-object
Filename: helpers/previewer_helper.php
Line Number: 211

This addresses #109.

### DIFF
--- a/application/helpers/previewer_helper.php
+++ b/application/helpers/previewer_helper.php
@@ -228,7 +228,10 @@ function format_authors($authors, $flags = 0, $max = 0)
 
 function format_author($author, $flags = 0)
 {
-	return format_authors(array($author), $flags);
+	if (isset($author))
+		return format_authors(array($author), $flags);
+	else
+		return '';
 }
 
 function format_playtime($seconds)


application/helpers/previewer_helper.php
```diff
@@ -228,7 +228,10 @@ function format_authors($authors, $flags = 0, $max = 0)
 
 function format_author($author, $flags = 0)
 {
-	return format_authors(array($author), $flags);
+	if (isset($author))
```
notartom, line 231: This should work, but I think we need something on L148 as well.

notartom (changes requested):

application/helpers/previewer_helper.php
kgroeneveld: Currently the code at L148 (in function `format_author_name()`) is only ever called from `format_authors()`. When that function is called directly I would expect the array to just be empty if there are no authors such that `format_author_name()` will not get called. Although I can't say I am absolutely 100% confident about that...

The definite case we are seeing an issue with now is when null is passed into `format_author()` which then passes an array with one null element to `format_authors()` for which the added `isset` in `format_author()` is sufficient.

notartom: Ah, right you are, apologies for the rushed review.

PR #113, opened 2021-04-06
API: only include `url_other` if set

### LINKED ISSUE #112: PHP errors on API calls
> Most API calls which I was using as recently as YESTERDAY without encountering errors, are now failing:
> 
> Example 1: Detecting what the End Of Book Data will look like:
> https://librivox.org/api/feed/audiobooks/?offset=15500&limit=1000&fields={id}
> 
> Example 2: simple call
> https://librivox.org/api/feed/audiobooks/?&offset=49
> 
> Both of these calls and a significant number of others yield:
>
> A PHP Error was encountered
> Severity: Notice
> Message: Undefined index: url_other
> Filename: libraries/Librivox_API.php
> Line Number: 142

notartom:
Previously, we'd blindly read $row['url_other'], even in cases where
it wasn't set. With this patch, we check with isset first.

Resolves #112

### DIFF
--- a/application/libraries/Librivox_API.php
+++ b/application/libraries/Librivox_API.php
@@ -139,7 +139,7 @@ class Librivox_API{
 			$project['url_project'] 	= $row['url_project'];
 			$project['url_librivox'] 	= $row['url_librivox'];
 
-			$project['url_other'] 		= $row['url_other'];
+			$project['url_other'] 		= isset($row['url_other']) ? $row['url_other'] : '';
 
 			$project['totaltime'] 		= $row['totaltime'];
 			$project['totaltimesecs'] 	= time_string_to_secs($row['totaltime']);


kgroeneveld:
The proposed patch does fix the specific examples in #112.

At random I tried https://librivox.org/api/feed/audiobooks?offset=100 on my test server and it gave a different PHP error:
>Severity: Notice
Message:  A non well formed numeric value encountered
Filename: helpers/general_functions_helper.php
Line Number: 185

I keep wondering if we should be adding these `isset` calls more on the source side of this data. For example, set the `url_other` element to empty string as soon as it is read from the database. It seems that would result in less `isset` calls all over the place. I am not sure what is normal in a typical PHP program for handling this type of issue. And making that change now could have unexpected results unless you track down every use of `url_other`. For now it is probably safer to just patch it as your proposed patch has it.

kgroeneveld:
I think the "A non well formed numeric value encountered" error is a bit of a false alarm, for now. It only seems to happen with PHP7 but is OK with PHP5.6. PHP7 does not like the space at the end of the `totaltime` field that is in some projects.

notartom:
> I keep wondering if we should be adding these `isset` calls more on the source side of this data. For example, set the `url_other` element to empty string as soon as it is read from the database. It seems that would result in less `isset` calls all over the place. I am not sure what is normal in a typical PHP program for handling this type of issue. And making that change now could have unexpected results unless you track down every use of `url_other`. For now it is probably safer to just patch it as your proposed patch has it.

Agreed, `isset`s all over the place are an ugly hack, but also agreed that they're better than nothing until we implement something more wholesale.

PR #116, opened 2021-04-07
Stop using max_password_length config option

### LINKED ISSUE #115: Unable to reset password
There seems to be a problem with the password reset in the workflow since the upgrade. Entering a password with 11 characters results in the following error:

Change Password
The New Password field can not exceed characters in length.

notartom:
Ion Auth 3 has deprecated the `max_password_length` config option (and
rightfylly so). Remove any references to it in our code. This fixes an
issue where password could not be changed or reset because they had to
be shorter than a `max_password_length` value that was not set.

Resolves #115


twinkietoes-on:
This fix (what you did on the fly on the live site) has fixed the issue.

PR #120, opened 2021-04-16
WIP: Ignore mp3 files that mp3gain cannot process

### LINKED ISSUE #119: Improve handling of incorrect paths in validator
If a path is incorrect in the validator, this is not handled gracefully, and instead we get a whole bunch of errors because the following code in `application/libraries/Librivox_mp3gain.php` that parses the `mp3gain` output does not have a second line to work with:

```php
$this->flag     = ' -o  ';
$command = $this->mp3gain .  ' ' . $this->flag . $dir. $file_name;
exec($command, $output);
$parts = explode("\t", $output[1]);
```

Here's a screenshot of an incorrect validator path (notice the period at the end of the first file):

Screenshot from 2021-04-10 14-56-14

And here are the PHP errors we get:

```
A PHP Error was encountered
Severity: Notice
Message: Undefined offset: 1
Filename: libraries/Librivox_mp3gain.php
Line Number: 73
```

```
A PHP Error was encountered
Severity: Notice
Message: Undefined index: playtime_string
Filename: libraries/Librivox_id3tag.php
Line Number: 171
```

```
A PHP Error was encountered
Severity: Notice
Message: Undefined offset: 2
Filename: libraries/Librivox_id3tag.php
Line Number: 172
```

```
A PHP Error was encountered
Severity: Notice
Message: Undefined index: audio
Filename: libraries/Librivox_id3tag.php
Line Number: 173
```

```
A PHP Error was encountered
Severity: Notice
Message: Undefined index: audio
Filename: libraries/Librivox_id3tag.php
Line Number: 177
```

```
A PHP Error was encountered
Severity: Notice
Message: Undefined index: audio
Filename: libraries/Librivox_id3tag.php
Line Number: 179
```

notartom:
In the validator, sometimes audio files either have a typo in their
path, or aren't actually audio files (when this patch was being
tested, the offending files were HTML files from a 301 redirect
response). In those cases, mp3gain does not return any useful output.
Specifically, instead of printing the header line plus the actually
useful line, it just prints the header line, with an error message
going to stderr (which we ignore because we only look at stdout).

This patch just makes the validator skip any such files.

Resolves #119

### DIFF
--- a/application/libraries/Librivox_id3tag.php
+++ b/application/libraries/Librivox_id3tag.php
@@ -161,6 +161,7 @@ class Librivox_id3tag{
 
                 //analyze volume
                 $mp3gain_result = $this->librivox_mp3gain->analyze_file($dir, $file);
+                if (! $mp3gain_result) continue;
 
 				$file_array = $this-> _create_file_array($freeze); //prototype
 
--- a/application/libraries/Librivox_mp3gain.php
+++ b/application/libraries/Librivox_mp3gain.php
@@ -68,9 +68,13 @@ class Librivox_mp3gain
 		$command = $this->mp3gain .  ' ' . $this->flag . $dir. $file_name;
 		exec($command, $output);
 
-		$parts = explode("\t", $output[1]);
+        if (count($output) > 1) {
+            $parts = explode("\t", $output[1]);
+            return $parts;
+        } else {
+            return false;
+        }
 
-		return $parts;
 
 	}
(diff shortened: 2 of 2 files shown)


notartom:
https://github.com/LibriVox/librivox-catalog/commit/7f63d1c0524c93a73d73f905dd076f4ea7ab41f3 has superceded this, closing.

PR #128, opened 2021-11-21
Make German project template consistent

### LINKED ISSUE #125: German Template Edits
the disclaimers in the german solo-project are different.

Disclaimer of the first section:

> "Abschnitt [i]#[/i] von P'älzische G'schichte' in der Mundart erzählt. Dies ist eine LibriVox Aufnahme. Alle LibriVox Aufnahmen [b]**sind lizenzfrei und in öffentlichem Besitz**[/b]. Weitere Informationen und Hinweise zur Beteiligung an diesem Projekt gibt es auf librivox PUNKT org." [Wenn Du magst: "Gelesen von [i]Dein Name[/i]."] "P'älzische G'schichte' in der Mundart erzählt, von Franz von Kobell. [i]Titel des Abschnitts[/i]."

Disclaimer for the other sections:

> "Abschnitt [i]#[/i] von P'älzische G'schichte' in der Mundart erzählt, von Franz von Kobell. Diese LibriVox Aufnahme ist [b]**in the Public Domain**[/b]." [Wenn Du magst: "Gelesen von [i]Dein Name[/i]."] "[i]Titel des Abschnitts[/i]"

The first is in german, the second in englich. The second should also be [b]**lizenzfrei und in öffentlichem Besitz**[/b].

By the way, there is one point in the templates that sounds not so good.

Section from book from author - translated as - Abschnitt **von** Buch **von** Autor.
If you think about the fact that there are German authors with a von in their name (Joseph **von** Eichendorff, Johann Wolfgang **von** Goethe), then there are 3 von in quick succession.
My suggestion:
Abschnitt **aus** Buch **von** Autor - translated - Section out of book from author.

What do you think about it?

Greetings from Germany.

ITD88:
resolves #125

Edit: I'm sorry, github confuses me so much. I wanted to create a second pull request for a separate issue but now it has put both together.

The second issue is that the uploader gives one filesize when you start uploading and a much small filesize when you finish. This is because [this file](https://github.com/LibriVox/librivox-catalog/blob/master/system/libraries/Upload.php#L496) is converting the filesize to kb after upload and that size is being interpretted as bytes. Multiplying by 1024 reverses this.


notartom:
Yeah, the GitHub pull request is weird to me as well. You can think of it as asking to merge your branch into master, so whatever commits you add to your branch will become part of the pull request. I think the generally accepted workflow is to have one branch per topic, or issue to fix, and then open one pull request for each branch.

In any case, thanks for putting in the effort! I've taken the liberty of rewording your commits (and fixing the template.php one, you missed multiplying by 1024 in the non-error part of the code), and pushing them directly as 5f19782 and 3d0c92f).

PR #130, opened 2021-11-23
Make Brief Summary By less strict

### LINKED ISSUE #83: Change edit ability in workflow profiles
In the workflow, a user's profile page looks like this:
profile

All fields except the User Name are currently editable by the user.

We would like to make the Display Name (which, in another issue, we would like renamed), Website, and Max Project Count only editable to admins. The average user should only be able to change their email address and password.

Low priority (4-5).

### LINKED ISSUE #123: "Brief Summary By" field parameters too strict
"When I submit the form, it gives me the following error and also deletes the recorded language choice and the selected author."

    The Brief Summary By field may only contain alpha-numeric characters, underscores, spaces and dashes.

Sometimes other characters are used (parentheses, slashes, etc.) so this needs to be less strict.

ITD88:
resolves #123 
resolves #83


kgroeneveld:
Ideally I don't think completely unrelated items should be included in the same PR.  And ideally the commit messages should be more detailed about what they are doing and which issues it fixes.

I tested the fix for #123 in this PR and it seems OK.

I also tested the fix for #83 and it has issues. Technically it does implement what is described in #83. But when it was asked to only allow admins to to change these fields what was really meant was MCs (and admins).  The MCs often refer to themselves as the admins but technically their user accounts are not admin accounts.

I just created #135 which has the same changes for #123 as this PR contains so that it can be considered separately. (Technically I also included an unrelated change, but it is at least somewhat related.)

I started working on #83 a long time ago but never finished. Now that it is fresh in memory again I hope to actually finish my change and create a new PR within the next couple days.

notartom:
@ITD88 has mentioned that they're still learning the GitHub PR workflow. I wanted to extract the useful stuff out of this PR into separate commits, a bit like I did with https://github.com/LibriVox/librivox-catalog/commit/5f197826871c81905281b2ad81dbee3a3acc7556 and https://github.com/LibriVox/librivox-catalog/commit/3d0c92f9420fb51562077433babaf97c8468a7ab from https://github.com/LibriVox/librivox-catalog/pull/128, but thankfully instead of sitting forgotten in my backlog, @kgroeneveld went ahead and actually did it. We can leave this open for now, and let's close it once the "equivalent" PRs have landed.

notartom:
I think with PR #137 this can be closed, right?

kgroeneveld:
> I think with PR #137 this can be closed, right?

Yes.  All the changes in this PR have now been merged as separate commits or implemented in alternate ways.  Since you already closed this PR it seems you already came to the same conclusion.

PR #134, opened 2021-12-29
Add Ukrainian, update German and Russian languages.

kgroeneveld:


notartom (approved):
As I speak neither German nor Ukrainian, I have to trust this blindly. The Russian bits look good, though ;)

PR #135, opened 2021-12-31
Relax "Brief Summary By" field validation.

kgroeneveld:
This PR contains one of the fixes from #130 to relax "Brief Summary By" field validation. It also removes the `alpha_dash_space` rule implementation completely as it is not really needed.


kgroeneveld:
Fixes #123.

application/libraries/MY_Form_validation.php
```diff
@@ -2,9 +2,4 @@
 
 class MY_Form_validation extends CI_Form_validation
```
notartom, line 3: This is the only function that this file contains, could we not disappear the entire file?

notartom (changes requested):

application/libraries/MY_Form_validation.php
kgroeneveld: I did consider deleting the entire file. But I thought it might be better to leave it there in case we ever wanted to re-add any custom logic to it. Although the history is always in git (if we think to look when if it is needed). I was also not sure if simply deleting the file would cause any issues or require any changes any where else.

I did just try deleting the file and everything does still seem to work okay. My Code Igniter knowledge is not enough to even fully understand how that file gets loaded.  I think there is logic to automatically look for MY_xxx classes to override the standard ones (configure by `$config['subclass_prefix'] = 'MY_';`).

So I am still on the fence whether it is a good idea to delete the file. But your suggestion to remove it probably pushes me over the edge. I will update the PR.

notartom (approved):

PR #137, opened 2021-12-31
Restrict fields that users can change in their profile.

### LINKED ISSUE #83: Change edit ability in workflow profiles

kgroeneveld:
Only an admin can change User Name (as it was before). Only admins or MCs
can change Display Name, Website and Max Project Count.

Fixes #83.

### DIFF
--- a/application/controllers/private/User.php
+++ b/application/controllers/private/User.php
@@ -94,7 +94,7 @@ class User extends Private_Controller
 		$this->ion_auth->update($user_id, $update);
 		$this->_update_groups($fields['groups'], $user_id);
 
-		$this->ajax_output(array('message' => 'Updated'), true);
+		$this->ajax_output(array('message' => 'Updated', 'user_id' => $user_id, 'update' => $update), true);
 	}
 
 	public function add_profile()
--- a/application/controllers/private/User.php
+++ b/application/controllers/private/User.php
@@ -44,40 +44,55 @@ class User extends Private_Controller
 
 	public function update_profile()
 	{
+		$this->load->helper('array');
+
 		$fields = $this->input->post(null, true);
 		$user_id = (empty($fields['user_id'])) ? $this->data['user_id'] : $fields['user_id'];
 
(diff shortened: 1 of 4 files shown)


notartom (changes requested):
I'm being nitpicky, but out of principle, can we at least add a comment explaining that simply disabling a form field doesn't actually prevent someone ill-intentioned from sending a specially crafted POST request to change a disabled field? I realize this is extremely unlikely to happen in practice, so the effort to value ratio swings heavily in favour of this current approach of just disabling the form field client-side in the brower without doing any server-side verification, but still... just a comment, out of principle :)

kgroeneveld:
I was actually working on some other issues with the profile update API such as if an MC updates another MC's profile then it always removes the MC role from that user (and it can only be added back by an admin). I was also planning on fixing the code to block the dis-allowed updates that you mentioned (and there are other existing examples).

Maybe I got a little ahead of my self on this PR and should finish the above changes and add them to this PR.  Then the comment you requested won't be needed.

kgroeneveld:
Its a good thing this did not get merged as is. After spending a bunch of time trying to improve the APIs that this screen uses and trying to actually exercise all possible cases I have found a couple bugs triggered by the original proposed change. The form validation fails on the disabled fields. And when a BC adds a reader these fields should still be enabled.

Its amazing how far down the rabbit hole this stuff can go even when the original request seems so simple. I thought I was very close to submitting an updated pull request when I ran into that last mentioned issue during my testing. And I have thought I was close a few times before that too...

kgroeneveld:
The more testing I did the more issues and rough edges I found with the user profile stuff.  Hopefully I didn't create any new bugs. @twinkietoes-on and I both tested my latest changes.

>I'm being nitpicky

Probably a lot more possible stuff in these commits to be nitpicky about now.  ;)

notartom (approved):

PR #145, opened 2022-10-24
No SQL injection in Poject_model.php

notartom:


arthurdarcet (approved):

notartom:
@kgroeneveld @garethsime I'd appreciate it if you folks can sanity check this for me. I'll do some testing in the staging environment (which is publicly accessible [1], so you can too), but I hate changing this kind of "core" stuff without unit tests or something similar to back us up, so I'm falling back on reviews and manual testing.

[1] https://forum.librivox.org/viewtopic.php?p=1456609&hilit=how+to+use+the+staging#p1456609

garethsime:
Thanks for doing this! My family and I have COVID, so I won't be able to make time to review this properly this week, sorry

I agree that the lack of automated testing is a bit scary. Are there any examples in the project of how to write tests that involve hitting an actual database? (Apart from my PR the other day, I haven't done anything in PHP for almost a decade now, so I don't know what the done thing is 🙂)

notartom:
I'm a Python guy myself, so I have no idea what the currently accepted best practices for PHP are. They're probably not hard to find and understand. Adding them to this codebase, however, is a different matter.

Don't sweat the timing thing, take the time you need to get better. It's been sitting here for months if not years, what's another few weeks :)

application/models/Project_model.php
```diff
@@ -561,29 +565,26 @@ function get_projects_by_reader_status($params)
...
+        $sql .= sprintf(
+            'UNION
+            SELECT p.id, p.project_type, p.title_prefix, CONCAT(s.title, " (in ", p.title, ")") as title, p.url_librivox, p.status, p.coverart_thumbnail, p.zip_url, p.zip_size, l.two_letter_code
+            FROM sections s
+            JOIN authors a ON (a.id = s.author_id)
+            JOIN projects p ON (p.id = s.project_id)
+            JOIN languages l ON (l.id = p.language_id)
+            WHERE s.author_id = %d %s AND p.is_compilation = 1
+            ORDER BY 4 ASC',
+            $params['authod_id'], $status_where);
```
arthurdarcet, line 585:
```suggestion
            $params['author_id'], $status_where);
```

arthurdarcet (commented):
using `sprintf` won't protect against sql injection - you would need to switch to `?` everywhere instead: all user-provided input should only be added to a query in the second argument to `$sql->query(…, …)`

application/models/Project_model.php
notartom: Right, I understand the concept. I'm assuming sprintf will throw an error if you try to feed a non-decimal to %d. Wherever I use %s the replacing string isn't user-provided.

But yeah, I'm not sure why I didn't use ? in this case for author_id, will fix.

arthurdarcet: ah yep, the `%d` is a nice quick fix, and I _think_ it closes most loopholes. It won't throw an error though, because php. 😬  https://onlinephp.io/c/6a733

application/models/Project_model.php
```diff
@@ -70,50 +64,57 @@ function get_projects($params)
...
+            WHERE p.status = "complete"
+            ORDER BY p.date_catalog DESC
+            LIMIT 0, %d', $limit);
 
         $query = $this->db->query($sql);
-        return $query->result();                
+        return $query->result();
     }
 
     function get_projects_summary($user_id, $status='active')
```
garethsime, line 79: I think this function isn't used anywhere, the only references I can find are in `controllers/private/Stats`, and they're commented out. Perhaps the better fix here is to just remove the function?

application/models/Project_model.php
garethsime, line 1: Do we also need to do `_search_all_terms`? (It looks like the code is still reachable via `controllers/private/Projects#ajax_search_catalog`.)

garethsime (commented):
Sorry about the long wait!

application/models/Project_model.php
notartom: Yep, done.

application/models/Project_model.php
notartom: Huh, that's a good find. It's a weird function, and I think I can just inline it (and improve it, by making it use $this->db->like, which is commented out right above).

application/models/Project_model.php
```diff
@@ -70,50 +64,57 @@ function get_projects($params)
...
-                WHERE p.status = "complete" 
-                ORDER BY p.date_catalog DESC 
-                LIMIT 0, '. $limit ;
+        $sql = sprintf(
+            'SELECT p.id, p.title_prefix, p.title, p.url_librivox, p.description, p.project_type, p.coverart_thumbnail, p.date_catalog, COALESCE(l.two_letter_code, l.three_letter_code) AS language_code, language
+            FROM projects p
+            JOIN languages l ON (l.id = p.language_id)
+            WHERE p.status = "complete"
+            ORDER BY p.date_catalog DESC
+            LIMIT 0, %d', $limit);
```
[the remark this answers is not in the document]
notartom: Done. Why `array($limit)` though instead of just `$limit`?

notartom:
As far as I can tell I've tested all the changed code here, and it's working. I can leave this open for another week or two if anyone has the energy to review it again, but to me at least, this can go in.

application/models/Project_model.php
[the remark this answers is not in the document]
garethsime: https://codeigniter.com/userguide3/database/queries.html#query-bindings

I was just following the example in the docs, no other reason 🙂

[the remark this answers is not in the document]
notartom: Doh, of course, it's an array for the general case of multiple bindings.

application/controllers/api/Feed.php
```diff
@@ -86,10 +86,10 @@ public function authors_get()
 		}
 	}
 
-	public function lastest_releases_get()
+	public function latest_releases_get()
```
garethsime, line 89: This changes the API path, right?

Before: `https://librivox.org/api/feed/lastest_releases`
After: `https://librivox.org/api/feed/latest_releases`

(I don't know who uses that API, sorry.)

garethsime (commented):
I'm sorry I've been pretty slow on the reviews, this is a good one to get done, so thanks for putting in the time :)

application/controllers/api/Feed.php
notartom: Yep, good catch. I had to go back and add https://github.com/LibriVox/librivox-catalog/commit/148ab02ee6b51550b8b6c643464bd46878244cb5 to keep backwards compatibility with whatever was using that misspelled endpoint in the wild.

PR #150, opened 2022-12-18
Update copyright warning strings with new Canadian rules.

kgroeneveld:
Starting in 2023 Canada is changing to death plus 70 years instead of 50 years.


notartom:
I can only vouch for the French, Russian and English strings, but I'm going to assume they've all been provided/checked by folks fluent in that language.

PR #157, opened 2023-02-09
Remove unused variables from advanced search

### LINKED ISSUE #155: Searching the catalog by reader is slow
Hi! I'm currently running the latest stuff from librivox-ansible locally, which is using librivox-catalog `2d270e76f8bfbb0e9f4e75a68c06d3e81ae2a799` from 30th December.

When I run the catalog locally and do a reader-only search, it takes about 16s to respond. I noticed #127, which looks like it solves this problem, but it's still slow for me both locally and when I run the same queries against the production site.

### Expectation

Searching the catalog by just reader returns results reasonably quickly.

### Actual

Searching the catalog by just reader returns results after about 16 seconds.

### What I've tried so far

It looks like it takes about 16s to respond locally - it takes about 8 seconds per call to `Librivox_search#advanced_title_search`, of which there are two (one for the first page of results, and one to decide how many pages there are in total).

I noticed that the vast majority of the time is spent on fetching projects by section author when building `$reader_clause`:

```php
$reader_clause = '';
if (!empty($params['reader']))
{
  ...
  $section_project_ids = $this->_get_projects_by_section_author($params['author']);
  ...
}
```

But that variable (and `$section_reader_clause`) is never used, so it's sucking up a lot of time doing not much. (It looks like it might've been copy/pasted from building the author clause.) When I remove that call, the query loads in ~320ms rather than 16s.

Including an author in the search also gives good performance since `$params['author']` isn't empty then, so we don't try and do big joins and string operations on every single author in the database.

I'm running all of this locally using Docker rather than VirtualBox, but the numbers seem comparable to what I get if I do the same queries against the production site. If you agree that the `$section_project_ids` variable isn't being used and is sucking up a lot of time, then I'm happy to raise a PR to remove it :slightly_smiling_face: 

Thanks!

garethsime:
Fixes #155.

Calling `$this->_get_projects_by_section_author($params['author']);` without an author ID is really expensive, and we just throw the result away anyway. I also removed a few other unused variables while I was there, though the performance impact of those was negligible.

This should make searching by reader without an author much faster.

### DIFF
--- a/application/libraries/Librivox_search.php
+++ b/application/libraries/Librivox_search.php
@@ -141,20 +141,9 @@ class Librivox_search{
 		$reader_clause = '';
 		if (!empty($params['reader']))
 		{
-
 			$exact_match = (isset($params['exact_match'])) ? true: false;
-			//var_dump($exact_match);
-
 			$project_ids = $this->_get_projects_by_reader($params['reader'], $exact_match);
-			$section_project_ids = $this->_get_projects_by_section_author($params['author']);
-
 			$reader_clause = ' AND p.id IN (' . implode(',', $project_ids) . ') ';
-
-			$reader_ids = $this->_get_reader_ids($params['reader'], $exact_match);
-			$section_reader_clause = ' AND ( st.source_id IN (' . implode(',', $project_ids) . ') AND st.section_reader_id IN  (' . implode(',', $reader_ids) . ') )  ';
-
-
-
 		}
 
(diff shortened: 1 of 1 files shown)


notartom:
This code predates me by a long time, and the only person who understood is has long since moved on to other things, but both code inspection and testing bear this out - it's unused, and it makes search by reader much faster. Thanks!

Issue #132, opened 2021-12-21
Add language, other data to the Push to Archive during cataloging


twinkietoes-on:
A user suggested:
> Language tags at IA are not working: On https://archive.org/details/librivoxaudio, it says there are 16,244 results, but only 69 items in English, 2 in German, and 1 in French.  It is not possible to have a list of all LibriVox recordings [at Archive] in a given language.

This would require the language data to be sent when the project is uploaded to IA during cataloging. I doubt we could make it retroactive, but it may be nice to have new projects bring over the language data.

At the same time, it would be nice to bring over the total project duration also, if possible. This is populated in the project database sometime during the "upload to Archive" step. It would be nice to plug that info into the "Run Time" field at Archive as well.

Priority for both: 5 (low)

hornc:
Relates to / (near) duplicate of #8

hornc:
I have started looking into these issues and the archive.org metadata for the LibriVox collection -- `runtime` is populated automatically when things work well, as part of the audio file processing and format generation. 17,650 items in the collection have a `runtime`  Only 310 items don't, and I'm not yet sure why.

twinkietoes-on:
Are you speaking of the LibriVox catalog, where some run times are missing? 
That happens usually by human error (the cataloging system populates that field when sending the files to Archive, but if the MC has a browser window open with the database entry and makes an edit there, the run time and some other info auto-populated upon Archive delivery are erased, since they weren't in the database entry in their browser window. 
As far as I know, they don't get added to the Archive pages unless the MC does it manually.

hornc:
@twinkietoes-on Thanks for the response. I was talking about the archive.org metadata and collection. I was going by the statement from https://archive.org/developers/metadata-schema/index.html#runtime

runtime:
> usage notes: Uploader can set this field, but most often we have determined and set this value during the derive process.

Which seems plausible (but I don't know for sure if it's true), and the fact that most of the collection on archive.org has a runtime: https://archive.org/search?query=collection%3Alibrivoxaudio+AND+runtime%3A*  vs. the ~310 which don't: https://archive.org/search?query=collection%3Alibrivoxaudio+AND+NOT+runtime%3A*

I can also imagine any auto-generation on the archive.org side being over-ridden in some cases  side, if a blank runtime is explicitly passed or something, so maybe it's complicated.

[...]

PR #161, opened 2023-04-03
Add language to ia metadata

### LINKED ISSUE #8: Include language metadata in Internet Archive uploads
I noticed that all but a handful of LibriVox files on archive.org only mention the language in the description, but don't include the corresponding metadata field. That means that e.g. filtering for a specific language on the Internet Archive doesn't work.

I think this can be solved by passing the `three_letter_code` for the language as a parameter down [here](https://github.com/LibriVox/librivox-catalog/blob/9836940e68c36237698c5df242a3cb3d4fd4ebd5/application/libraries/Iarchive_uploader.php#L70) and then adding it as `x-archive-meta-language` to the header.

Unfortunately I don't have any experience writing PHP, otherwise I'd just send a PR to make this change.

hornc:
closes #8
and partially #132 

I'd like it if there was a 3 letter language code rather than a full name. (Hopefully done now with 42f73de8d8dfd3ce25bc29834c6cfe84647747ac)

Some of the existing code suggests there is a `3_letter_code`  (https://github.com/LibriVox/librivox-catalog/search?q=three_letter_code) , but I'm not completely sure whether this is guaranteed to exist, or where it comes from.

The archive.org fields to target are:
* https://archive.org/developers/metadata-schema/index.html#language
* https://archive.org/developers/metadata-schema/index.html#runtime

If I can determine a good source -> destination mapping, I can very likely make retrospective updates to the existing archive.org items happen.

I have not tested this code yet (not sure the best way to do so).

### DIFF
--- a/application/controllers/private/Iarchive_upload.php
+++ b/application/controllers/private/Iarchive_upload.php
@@ -48,7 +48,7 @@ class Iarchive_upload extends Private_Controller
 
 		$description = $this->_get_full_description($params, $project);
 		$params['description'] = trim(preg_replace('/\s+/', ' ', $description));  //trims all newlines before placing in header
-		$params['language'] = $this->data['language'];
+		$params['language'] = $this->data['language_code'];
 
 		// Close db connection before uploading to avoid hogging connections
 		$this->db->close();
@@ -109,6 +109,7 @@ class Iarchive_upload extends Private_Controller
 		$language = $this->language_model->get($project->language_id);
 
 		$this->data['language'] = empty($language) ? 'English' : $language->language;
+		$this->data['language_code'] = empty($language) ? 'eng' : $language->three_letter_code;
 
 		$this->load->model('project_model');
 		$this->data['readers'] = $this->project_model->create_project_reader_list($project->id);
--- a/application/controllers/private/Iarchive_upload.php
+++ b/application/controllers/private/Iarchive_upload.php
@@ -48,6 +48,7 @@ class Iarchive_upload extends Private_Controller
 
 		$description = $this->_get_full_description($params, $project);
 		$params['description'] = trim(preg_replace('/\s+/', ' ', $description));  //trims all newlines before placing in header
+		$params['language'] = $this->data['language'];
 
 		// Close db connection before uploading to avoid hogging connections
 		$this->db->close();
--- a/application/libraries/Iarchive_uploader.php
+++ b/application/libraries/Iarchive_uploader.php
@@ -63,13 +63,14 @@ class Iarchive_uploader
 		$headers[] = 'x-archive-meta-title:' . $title;
 
 		//additional info
-		$headers[] = 'x-archive-meta-creator:'.$params['creator'] ;
+		$headers[] = 'x-archive-meta-creator:'.$params['creator'];
 		$headers[] = 'x-archive-meta-description:'.$params['description'];
-		$headers[] = 'x-archive-meta-date:'.$params['date'] ;
-		$headers[] = 'x-archive-meta-subject:'.$params['subject'] ;
+		$headers[] = 'x-archive-meta-language:'.$params['language'];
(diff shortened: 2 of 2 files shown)


notartom:
So this looks correct, and I could deploy this on the staging server. @twinkietoes-on, if I deploy this on the staging server, would you be able to try uploading a few test files, to see if the language correctly appears in the metadata of the uploaded items?

twinkietoes-on:
I'll have to remind myself how to do it. I'll BCC Jo on this email; maybe she will help test.

    On Saturday, April 15, 2023 at 11:41:46 a.m. EDT, Artom Lifshitz ***@***.***> wrote:  
 
 


So this looks correct, and I could deploy this on the staging server. @twinkietoes-on, if I deploy this on the staging server, would you be able to try uploading a few test files, to see if the language correctly appears in the metadata of the uploaded items?

twinkietoes-on:
I got into the test server and uploaded this project (saying it was Spanish): https://archive.org/details/joy_other_poems_2304_librivox
I don't see the language info having been imported.

hornc:
@twinkietoes-on thank you for testing this! I don't see the `language` field there in the item either, or on the task that was submitted. I'll have a look to see whether there is a problem with the archive.org API payload.

notartom:
Lemme play arrive with the code on the staging server, now that I have a project to use for testing, and have figured out where the "upload to archive.org" button is :)

twinkietoes-on:
I'm going to delete that test project from Archive, unless anyone has an objection. Let me know before 8 pm Eastern if you do (in 32 minutes)

    On Tuesday, April 18, 2023 at 07:05:16 p.m. EDT, Charles Horn ***@***.***> wrote:  
 
 


https://archive.org/download/joy_other_poems_2304_librivox/joy_other_poems_2304_librivox_meta.xml Is where the metadata should appear, and it shows that it was updated with the recent date time: 2023-04-18 16:28:27. x-archive-meta-language: is the correct way to send the metadata using the ias3 API as documented here: https://archive.org/developers/ias3.html ... so I think the code is correct for this part.

For Spanish I would expect the 3-letter code spa to be sent. If something was wrong it might get set to the default eng, I'm surprised it's not being set at all though.

notartom:
OK, that was my bad, when @twinkietoes-on first tested in the staging environment, the wrong branch was checked out. When I attempted it again, I used the same project (https://archive.org/details/joy_other_poems_2304_librivox), and it looks like once uploaded, any further upload attempts do not change the metadata. I then tried again with https://archive.org/details/joy_other_poems_2304_librivox_second and https://archive.org/details/joy_other_poems_2304_librivox_third, and in both cases the language shows up correctly, so I think we're good.

@twinkietoes-on can you please delete all those testing projects from archive?

twinkietoes-on:
Will do!

    On Tuesday, April 18, 2023 at 08:43:54 p.m. EDT, Artom Lifshitz ***@***.***> wrote:  
 
 


Merged #161 into master.

twinkietoes-on:
The German ones are showing up as the 3-letter code, deu. 
Example:
https://archive.org/details/reise_um_die_welt_erste_abt_2304_librivox

hornc:
@twinkietoes-on archive.org has flexible, but unfortunately somewhat inconsistent handling of language codes.  The preferred 3 letter code is from the [MARC language code list](https://www.loc.gov/marc/languages/language_code.html), but ISO 639-3 and ISO 639-2 are also generally supported.

`German`, `ger` and `deu` are all valid in the metadata, and the LibriVox audio collection language filter now finds all of these using the `language` filter: https://archive.org/details/librivoxaudio?and[]=languageSorter%3A%22German%22  so that's one usecase that has definitely improved.

`deu` does appear to be left out from the item details page display logic though, which isn't great.

My feeling is that archive.org should handle ISO 639-3 better, rather than librivox have to translate the codes. Librivox seems to be consistent with its ISO 3 letter language codes, which work well. Many other archive.org items have ISO 639-3 codes already.

PR #163, opened 2023-04-16
Add cover art to RSS feeds

garethsime:
Hi! I saw pull request #10 and I would like this feature too, so I thought I'd have a go reviving that since the PR seems a bit dead.

@kgroeneveld had some comments about it at the time:

> The coverart_jpg field is empty for some items in the catalog. Most other spots where the cover art is displayed will substitute a generic image in this case. I wonder if the RSS feed should do the same or if the RSS feed should just omit the image tag in this case? The pull request as is will result in an empty href in this case which doesn’t seem good.

I've changed the code to only render the `<itunes:image />` when the image is actually there. In the case where there is no cover art, I've just left it as-is and hopefully podcast apps do something sensible. (The app I'm using, Podcast Addict, just uses the book title with a random background colour if there's no art present, which is a nice default.)

> The Apple specs for the <itunes:image> tag (https://help.apple.com/itc/podcasts_connect/#/itcb54353390) specify the size must be minimum 1400x1400. I think most of the LibriVox images are a lot smaller than this. I am not sure how much that would matter.

The images I've seen are indeed smaller. I don't have an Apple device to test on, but it all works fine on Podcast Addict on Android, which is what I'm using. If I had to guess, I'd look at https://podcasters.apple.com/support/902-troubleshooting-artwork-issues, which seems to imply that the art just won't show if it doesn't meet the requirements? Hard to say

> I wonder if it would be better to use a standard RSS <image> instead of the itunes one? Or maybe include both? I don’t really have much experience with RSS...

I have no idea, sorry.

### DIFF
--- a/application/views/rss/rss.php
+++ b/application/views/rss/rss.php
@@ -18,6 +18,11 @@
   <itunes:category text="Arts">
     <itunes:category text="Literature" />
   </itunes:category>
+  <?php
+  if ($project->coverart_jpg) {
+    echo sprintf('<itunes:image href="%s" />', $project->coverart_jpg);
+  }
+  ?>
   <!-- file loop -->
   <?php foreach($sections as $section): ?>
   <item>


notartom:
Seems to work, and the W3C validator [1] is happy with it, so LGTM.

Couple of notes on potential further work through:
1. Should we also add cover images to the latest releases feed? (https://librivox.org/rss/latest_releases, from https://librivox.org/pages/librivox-feeds/)
2. There are some validation errors unrelated to cover images at [1], we should probably fix those.

[1] https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Flibrivox.org%2Frss%2F17623

garethsime:
Neat, I didn't know about that validator. I'll make sure to run things through it next time.

- I noticed afterwards that most other data is wrapped up in `<![CDATA[…]]>` to prevent any mangling of the XML (intentional or otherwise), so I might add that for the image too

- Adding it to other feeds sounds like a good idea

- Changing the category to Books seems easy to do, though I always worry about breaking people's setups when changing tags. Maybe we should have both the old and new one instead? ([Latest categories](https://podcasters.apple.com/support/1691-apple-podcasts-categories))

- Language and explicit tags would be good, but I don't know anything about either, so I'd have to look when I'm back at my computer

- I'd love to fix the duration for my own use as well, though I see there's already some code there, so it might not be so straightforward as just plonking in some preexisting data I guess

Thanks for reviewing!

notartom:
>     * Changing the category to Books seems easy to do, though I always worry about breaking people's setups when changing tags. Maybe we should have both the old and new one instead? ([Latest categories](https://podcasters.apple.com/support/1691-apple-podcasts-categories))

Yeah, even though the validator says this is wrong, it's "out there", which effectively makes it canon, so it's probably best to leave it as is.

Issue #32, opened 2021-01-12
Search result pagination improvement


twinkietoes-on:
current pagination at bottom of results:
  « 1 2 3 4 5 » (even if there are 200 pages).
 
requested pagination interface is:
 « 1 2 3 4 5 ... 778 » (Go page X)
 
when you are in the middle of results:
 « 1 ... 77 78 79 80 81 ... 778 » (go to page X)
 
where ...
 a) 778 is the whatever the last page Is.
 b) go to page (X) ... the (X) is an inputtable number.
 
in the case that an X is entered that is greater than the last page, just go to the last page.
 

 but crap:
 lots of requests for alphabetical results...
 
 wonder: can we put this on the Advanced Search Page as an additional filter...?
 
  Starting with (alphabet dropdown). 

Priority: 2

notartom:
I think we can close this now that https://github.com/LibriVox/librivox-catalog/pull/164 has landed?

twinkietoes-on:
I think so. A definite improvement - thanks, garethsime!

garethsime:
No worries 🙂 Notably though, the new work I've done doesn't include the 'Go to X' field. I'm not sure how far into the pagination people will want to go instead of doing another search though, so I don't know if it's worth doing

twinkietoes-on:
And if they're at all web-savvy, they can see they could simply change the page number in the URL. Ha!
    On Saturday, April 22, 2023 at 04:03:30 p.m. EDT, Gareth Sime ***@***.***> wrote:  
 
 


No worries 🙂 Notably though, the new work I've done doesn't include the 'Go to X' field. I'm not sure how far into the pagination people will want to go instead of doing another search though, so I don't know if it's worth doing

PR #164, opened 2023-04-17
Reworks the pagination formatting

garethsime:
See issue #32. This commit changes the pagination at the bottom of the page to better show how many pages are available. It doesn't include a "Go to X" field for jumping to a specific page.

I've also tweaked the requested format slightly - if there are fewer than ten pages, then we'll just show all of them.

Here's a demo video:

https://user-images.githubusercontent.com/5061320/232436456-ce596614-5b8a-43aa-8576-1f39bc4df3e2.mp4

### DIFF
--- a/application/core/Catalog_controller.php
+++ b/application/core/Catalog_controller.php
@@ -24,32 +24,76 @@ class Catalog_controller extends CI_Controller
 		$this->load->view($file, $this->data);
 	}
 
-	function _format_pagination($first_page, $page_count, $call_function = 'get_results')
+	function _format_pagination($current_page, $page_count, $call_function = 'get_results')
 	{
-		$pages = $this->_build_pagination_array($first_page, $page_count);
+        $page_count = intval($page_count); // It's a float, but we want ints. (Causes problems when generating ranges.)
 
-		$html = '<a class="page-number first" data-page_number="1" href="#" data-call_function="' . $call_function . '">&laquo;</a>';
+        // If it's less than ten pages, just render everything, otherwise, three either side will do
+        $num_links_either_side = $page_count < 10 ? 10 : 3;
+		$pages = $this->_build_pagination_array($current_page, $page_count, $num_links_either_side);
 
-		foreach ($pages as $key => $page)
+        $html = '';
+
+        // Print the link to the first page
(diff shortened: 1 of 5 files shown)


application/core/Catalog_controller.php
```diff
@@ -24,33 +24,77 @@ function _render($file)
...
 
-	function _format_pagination($first_page, $page_count, $call_function = 'get_results')
+	function _format_pagination($current_page, $page_count, $call_function = 'get_results')
 	{
-		$pages = $this->_build_pagination_array($first_page, $page_count);
-
-		$html = '<a class="page-number first" data-page_number="1" href="#" data-call_function="' . $call_function . '">&laquo;</a>';
-
-		foreach ($pages as $key => $page)
+        $page_count = intval($page_count); // It's a float, but we want ints. (Causes problems when generating ranges.)
```
garethsime, line 29: I wasn't sure what the PHP way of doing this is, sorry.

`_build_pagination_array` normally returns an array of integers, but if `$current_page` is near the end, then I think the `min(($first_page + $num_links), max($page_count, 1))` ends up being a float instead of an int, and so the returned range is a range of floats rather than a range of ints.

I didn't notice it until I tried to do a `===` on it to compare `$page` and `$page_count` :slightly_smiling_face:

notartom: I'm no PHP expert either, we'll just have to live with this weirdness.

application/core/Catalog_controller.php
```diff
@@ -24,33 +24,77 @@ function _render($file)
 		$this->load->view($file, $this->data);
 	}
 
-	function _format_pagination($first_page, $page_count, $call_function = 'get_results')
+	function _format_pagination($current_page, $page_count, $call_function = 'get_results')
 	{
-		$pages = $this->_build_pagination_array($first_page, $page_count);
-
-		$html = '<a class="page-number first" data-page_number="1" href="#" data-call_function="' . $call_function . '">&laquo;</a>';
```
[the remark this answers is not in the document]
notartom: 🤷 It's probably fine?

notartom:
Does what it says on the tin, and is a definite improvement to the pagination links. LGTM!

PR #166, opened 2023-04-24
Add Librivox project id to archive.org metadata as call_number

hornc:
Currently the archive.org metadata provides no direct way to get the original LibriVox project id.

There is usually a link to the LibriVox project slug (the name, e.g. https://librivox.org/sense-and-sensibility-by-jane-austen/ ) in the archive.org `description` field, but the format varies.

This PR adds the project_id to a standard archive.org field [`call_number`](https://archive.org/developers/metadata-schema/index.html#call-number) to enable automated references back to LibriVox, for checking or accessing the original metadata etc.

Example screenshot showing where the call_number appears on the archive.org details page (this one, 620,  was added manually to https://archive.org/details/0_sense_and_sensibility_librivox ):
**Screenshot:**

image

**Background:**
I'm currently working on generating MARC records for archive.org audiobooks, investigating what metadata is required and available. Being able to include a LibriVox URL (e.g. https://librivox.org/620 ) is helpful, and provides a way to locate any other metadata that happens to be missing from archive.org if it is ever required.

Perhaps I should create a MARC issue for this work. I notice MARC export is a documented 'feature', but it appears to be aspirational :)  My current code is standalone Python, but once I can get the algorithm working and the output verified, I could port it to PHP, or alternatively have the MARC records generated for archive.org as part of a separate process, and librivox-catalog can link to a standard location for each item. I'm still trying to confirm the MARC output is correct and sufficiently detailed to be useful to libraries, so it's a bit proof-of-concept at this stage.

### DIFF
--- a/application/controllers/private/Iarchive_upload.php
+++ b/application/controllers/private/Iarchive_upload.php
@@ -33,6 +33,7 @@ class Iarchive_upload extends Private_Controller
 		//I think we need to load these files one by one; I will investigate doing a single upload,
 		//but am worried about several technical issues
 
+		$params['project_id'] = $project->id;
 		$params['project_slug'] = $fields['upload_name'];
 		$params['title'] = $fields['upload_title'];
 		$params['file_location'] = $dir;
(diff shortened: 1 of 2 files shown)


notartom:
Looks to be working OK:

https://archive.org/details/joy_other_poems_2304_librivox_fifth_test

I guess it makes sense that no "magic" happens with that test, since it was uploaded from the staging environment, and that project ID on the production server isn't finished yet and isn't even the same book.

Issue #158, opened 2023-02-15
Reader page access ignores query string


twinkietoes-on:
(I hope this is clear enough. It's just beyond my thorough understanding. I'm copy/pasting.)

Any access to the reader page (eg, https://librivox.org/reader/8463) clobbers the querystring.
For example:
https://librivox.org/reader/8463?primary_key=8463&search_category=reader&search_page=4&search_form=get_results

The query includes the parameter "search_page=4", so the expected behavior would be to display results starting on page 4. Instead the search script ignores the entire querystring and displays the first page of results.

https://forum.librivox.org/viewtopic.php?p=2156227#p2156227

redrun45:
Thanks to garethsime (thank you!), we no longer lose the page on reader searches.  We might close this as fixed, since I don't see anything else there to be broken by the clobbering.

notartom:
So as @garethsime said in https://github.com/LibriVox/librivox-catalog/pull/167, his patch fixes Author, Group, and Reader pages. Are there other pages where the query string is being ignored that haven't been fixed by his patch?

redrun45:
Not that I'm aware of.  I think if we run into any more such issues, we can file them separately.

PR #167, opened 2023-04-30
Keep track of page numbers between refreshes

garethsime:
This PR should help with some of issue #158, at least for the Author, Group, and Reader pages. 

On a page of search results, we're tracking the current page number in memory in the browser to know where we are. We also push that value into the history under the query param `search_page`.

However, when a user browses away and comes back, we don't make use of it. (The parameter isn't propagated by the PHP controller and it's not checked by the Javascript.)

Originally, I pushed this PR with a change to make sure the PHP controllers are passing through the `search_page`, but I decided that it would be better to have the frontend read the value instead. (I've left both commits in the PR, but I can squash them if you want the history kept tidy.)

Notably, I'm using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams), which [isn't supported by every browser](https://caniuse.com/urlsearchparams), but is supported by most.

### DIFF
--- a/application/views/catalog/partials/footer.php
+++ b/application/views/catalog/partials/footer.php
@@ -15,6 +15,25 @@
 
 <script type="text/javascript">
 
+	function get_search_page_from_url() {
+		var query_string = window.location.href.slice(
+			window.location.href.indexOf('?')
+		);
+
+		if (typeof URLSearchParams !== 'function') {
+			// Manual parsing is annoying and most browsers support URLSearchParams
+			return 1;
+		}
+
+		var query_params = new URLSearchParams(query_string);
+		var search_page_param = parseInt(query_params.get('search_page'));
+		if (typeof search_page_param === 'number' && search_page_param > 0) {
+			return search_page_param;
+		} else {
+			return 1;
+		}
+	}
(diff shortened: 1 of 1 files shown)


notartom:
> Originally, I pushed this PR with a change to make sure the PHP controllers are passing through the search_page, but I decided that it would be better to have the frontend read the value instead. (I've left both commits in the PR, but I can squash them if you want the history kept tidy.)

Yes please :) I don't want to go all OCD on tidy commit history, but doing a thing in one commit, then undoing it immediately in the following just looks kinda silly.

garethsime:
Haha, I've never used Gerrit I'm afraid. When I raise a PR in Github, there's an option to allow maintainers to push changes to the branch: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I've got that one ticked I think, but I'm not sure if that includes permission to force-push to a branch, which you'd have to do to squash it. (Either way, it's a bit of a hassle for you to have to add my fork as a remote and pull the branch etc.)

Issue #154, opened 2023-01-04
Some characters now not allowed in author date fields


twinkietoes-on:
Trying to use a dash in the author date field (such as "ca. 1315-1325") or a slash (such as "1315/1325") results in the field not saving. Those are the only two characters discovered that are occasionally used.
This has only been an issue since the upgrades. It worked before.

twinkietoes-on:
We also had a possible rejection of a period in an author name field. So:
- non-numerical characters in author birth/death year fields
- non-numerical characters in copyright & edition year fields
- periods in author name fields (possibly)

notartom:
I believe it worked before the upgrade because the new version of the database software is more strict in what input it accepts (https://mariadb.com/kb/en/sql-mode/#strict_trans_tables). So previously, if a field in the database was listed as an integer, but the input was some alphanumeric string of characters, there would be some magic at the database level that would accept the input, possibly mangling it somewhat, or dropping that particular field altogether while still updating the other ones in the query (I'm not very knowledgeable about this, hence the handwavy "magic" explanation).

@garethsime's PR #169 is one way to address this, but it would mean author YOB and YOD would only be numbers. That would make sense to me, but I'm not sure if there's precedent for input other kinds of data into those fields.

garethsime:
Yeah, my PR definitely doesn't solve the underlying problem, it just gives a better error message. (There were a couple of threads in the forum where people were expressing frustration that they were just getting a generic error message with no information on how to fix it.)

twinkietoes-on:
There's definitely precedent for non-numeric characters in YOB and YOD fields: ca., BC or BCE, AD or CE, and fl. being the primary ones.

    On Saturday, September 9, 2023 at 05:29:46 p.m. EDT, Gareth Sime ***@***.***> wrote:  
 
 


Yeah, my PR definitely doesn't solve the underlying problem, it just gives a better error message. (There were a couple of threads in the forum where people were expressing frustration that they were just getting a generic error message with no information on how to fix it.)

notartom:
(from PR #169)
> For the more technical explanation -- When we save the form generation, that all goes into the form_generators table. Any new authors that are created as part of the generator are saved into form_generators_authors, which has the auth_yob and auth_yod columns as four-digit integers. This doesn't align well with the actual authors table, which allows pretty much anything (varchar(10)), and I have no idea how people specify AD vs. BCE using the generator (negative, maybe?).

[...]

PR #169, opened 2023-08-21
Form Generator - Require author and translator dates are numeric

garethsime:
This goes some way to addressing issue #154 -- users won't be able to enter non-numeric dates, which is unfortunate, but it does mean that they'll get a proper error message in the form instead of the random server error that they get right now.

For the more technical explanation -- When we save the form generation, that all goes into the `form_generators` table. Any new authors that are created as part of the generator are saved into `form_generators_authors`, which has the `auth_yob` and `auth_yod` columns as four-digit integers. This doesn't align well with the actual `authors` table, which allows pretty much anything (`varchar(10)`), and I have no idea how people specify AD vs. BCE using the generator (negative, maybe?).

I don't know enough about how the table is used to know whether we should consider migrating it to `varchar(10)` to match `authors`, but having the two consistent with each other would be good probably.

I also tried specifying the input fields as numeric in the HTML, but that didn't quite have the desired effect -- it would just submit the field as empty if it wasn't right, which meant we'd silently drop the data rather than get a validation error.

EDIT: I've added the `|numeric` to a few other fields that were missing it as well.

### DIFF
--- a/application/controllers/public/Project_launch.php
+++ b/application/controllers/public/Project_launch.php
@@ -209,15 +209,15 @@ class Project_launch extends Public_Controller
 		$this->form_validation->set_rules('auth_id[]', 'Author id', 'trim|xss_clean');
 		$this->form_validation->set_rules('auth_first_name[]', 'lang:proj_launch_auth_first_name', 'trim|xss_clean');
 		$this->form_validation->set_rules('auth_last_name[]', 'lang:proj_launch_auth_last_name', 'trim|xss_clean');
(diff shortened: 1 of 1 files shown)


notartom:
I think we need clarity on how those fields are intended to be used - see https://github.com/LibriVox/librivox-catalog/issues/154#issuecomment-1712624992.

notartom:
I've changed the schema to use `varchar(10)` for `form_generators_authors.author_yo(b|d)`, I think we can close this PR.

garethsime:
Awesome!

PR #174, opened 2023-11-19
Issue/#39

hamidsafdari:
Fix validator uploading one file at a time.


notartom (changes requested):
In my testing this didn't solve the issue.

On staging with this PR checked out, going to https://librivox.org/validator/18656, clicking 'Add new files' > 'Add files' > selecting multiple MP3 files on my computer > 'Start' on the top file in the list results in that one file uploading, the page refreshing, and the other files that were present disappearing.

Note that there's a workaround for this, as clicking on 'Start upload' in the main bar of buttons (as opposed to 'Start' next to an individual file) works as expected, and uploads all the files.

PS: Let's lose the .editotconfig file :)

hamidsafdari:
Stating the obvious here, but the changes were made to the JavaScript code and might need a hard refresh (Ctrl+F5). 
It also sometimes helps to manually clear the cache:

* Chrome: F12 > Application > Storage > Clear site data
* Firefox: Preferences > Privacy and Security > Cookies and Site data > Manage Data > Search for the website > Select and Click on Remove Selected.

notartom:
I did try a hard refresh, though I acknowledge that Firefox is sometimes weird with cache. I've had more predictable results with Chromium, so I can try again with that. Will report back.

hamidsafdari:
Also, I've kept this PR in draft mode because a refresh is needed after all the files are uploaded to render the file data. However, I do see some error when uploading some .mp3 files (unknown characters in filename and application/octet-stream as the MIME) but because the page reloads after all the files are uploaded, the errors are lost.

I intend to show a dialog to the user with the list of files that had a problem and reload the page when the user closes or clicks on OK.

hamidsafdari:
There is a callback called `stop` that Blueimp (the file upload library) calls after all uploads are done. I removed my old code and used that instead. Also, the errors are not passed to the `stop` callback. I had to collect them in a variable outside and use it in `stop`. 

To be consistent, I styled the upload result dialog after jQuery UI since it's also shown when editing the file fields. The alternative would be to use the Bootstrap modal styling.

With errors:
[validator-with-error.webm](https://github.com/LibriVox/librivox-catalog/assets/5158006/bb24bbfb-ddc2-4c01-9145-5e99849243f7)

Without errors:
[validator-with-no-error.webm](https://github.com/LibriVox/librivox-catalog/assets/5158006/d8d3c7ae-f23d-4e73-8a28-1e01a6c0bdc3)

notartom:
As mentioned in https://github.com/LibriVox/librivox-catalog/pull/175#issuecomment-1848828336 and https://github.com/LibriVox/librivox-catalog/pull/175#issuecomment-1848992764, looks like you've lost hope with my slow turnaround time for reviews and have moved on to bigger and better things. Because this is a fairly big PR, I'm not in a position to review it and make the adjustments myself. I'm going to close this. If you decide you want to come back, by all means reopen this PR.

PR #175, opened 2023-11-20
fix/#146 better error messages in uploader

hamidsafdari:
Detect if message is originated using AJAX and send appropriate response.

### DIFF
--- a/application/controllers/public/Uploader.php
+++ b/application/controllers/public/Uploader.php
@@ -18,8 +18,26 @@ class Uploader extends Public_Controller
 		ini_set('memory_limit', '128M');
 
 		$allowed_groups = array(PERMISSIONS_ADMIN, PERMISSIONS_BCS, PERMISSIONS_MCS, PERMISSIONS_UPLOADER, PERMISSIONS_PLS, PERMISSIONS_READERS);
-		if (!$this->librivox_auth->has_permission($allowed_groups, $this->data['user_id']))
+		$has_group_permission = $this->librivox_auth->has_permission($allowed_groups, $this->data['user_id']);
+		if (!$has_group_permission)
 		{
+			if (!$this->ion_auth->logged_in())
+			{
+				if (IS_AJAX)
+				{
+					$error = array('error' => 'Your session has been expired. Please login again.');
+					echo json_encode(array($error));
+					die();
+				}
+				redirect('auth/login');
+			}
+
+			if (IS_AJAX)
+			{
+				$error = array('error' => 'You don\'t have permissions for this area.');
+				echo json_encode(array($error));
+				die();
+			}
 			redirect('auth/error_no_permission');
 		}
 


application/controllers/public/Uploader.php
```diff
@@ -18,8 +18,26 @@ function __construct()
 		ini_set('memory_limit', '128M');
 
 		$allowed_groups = array(PERMISSIONS_ADMIN, PERMISSIONS_BCS, PERMISSIONS_MCS, PERMISSIONS_UPLOADER, PERMISSIONS_PLS, PERMISSIONS_READERS);
-		if (!$this->librivox_auth->has_permission($allowed_groups, $this->data['user_id']))
+		$has_group_permission = $this->librivox_auth->has_permission($allowed_groups, $this->data['user_id']);
+		if (!$has_group_permission)
```
notartom, line 22: Thanks for getting this started! It's a relatively small change, but it will definitely improve the UX.

I have a couple of suggestions to improve this, both to organize the code, and to improve what the user sees. I would imagine it something like this:

```php
# First determine the error condition, if any.
if (!$this->ion_auth->logged_in())
{
    $ajax_error = "Your session has expired, please login again.";
    $redirect_url = "auth/login";
}
else if (!$has_group_permission)
{
    $ajax_error = "You don't have permissions for this area.";
    $redirect_url = "auth/error_no_permission";
}

# Then, based on the presence of an error condition and whether we're being called by AJAX or not,
# echo or redirect.
if (IS_AJAX and $ajax_error)
{
    echo json_encode(array(
        'error' => $ajax_error,
        # Also set the name and size, to avoid 'undefined' and 'Nan' showing up on the web page
        'name' => '',
        'size' => ''));
}

if ($redirect_url)
{
    redirect($redirect_url);
}
```

notartom (changes requested):

notartom:
Oh, and if you tag the issue in your commit message, something like `Closes issue #175`, the issue will auto-close when this merges.

hamidsafdari:
I'm sorry, but I honestly felt discouraged by the response time on PRs and removed the project environment. I've also made other PRs #174, #176 and one in [librivox-ansible](https://github.com/LibriVox/librivox-ansible/pull/17). You have requested for change on those PRs too, and I updated my code, but there was no response after that.

notartom:
I get it. I've tried to set expectations that LibriVox reviews, especially when the PR is very big, seldom have quick turnaround times because other things in life take priority for me. I'll update this PR and merge it. If you're really no longer interested in working on the other PRs, I'll have to abandon them, as they will require a lot of review effort, and I don't want to put in the time if you've moved on to bigger and better things.

notartom:
Merged as https://github.com/LibriVox/librivox-catalog/commit/1c0a92179477e3589939da2cbd715f5c1942ff12

PR #179, opened 2023-12-30
Low-Code Fixes

### LINKED ISSUE #43: In keyword entry on project template, keep capitalization as entered
Currently the capital letters are converted into lower case at some point (when the project template data is submitted and the code generated?) Should stay with the case given it by the user.

Priority: 2

### LINKED ISSUE #44: Limit suggested validator directory name
Auto-generated Validator directory names that are too long do not get auto-deleted after cataloging. They're auto-generated using the project title. Shorten the suggested name somehow - truncate it after X characters?

Priority: 2

### LINKED ISSUE #74: In Project Template Generator, move the box for the different language choices in a more conspicuous spot
I see many newbies (and even old hands) recording in languages where we have a template for and they use the English one because they overlook that dropdown (right now it's in the top right corner) and are then told to change the template.
Maybe put it in the top middle before anything else to make it stand out?

The ideal scenario, IMHO, would be that when someone chooses the language of the work and submits the form, it would search for the template in that language and use it if it's there. But I don't know how hard something like that would be to implement.

Priority: 2

redrun45:
Well, there goes my attempt to keep the commit history clean.  I'm still getting the hang of git, let alone GitHub.  :neutral_face: 

Here's a collection of fixes with a relatively low code complexity.   I don't know if it's easier to accept these changes all from the same branch in one PR, or to keep them entirely separate.  For anything more than a very few lines of code, I'll definitely go for a separate PR.

Fixes #44 - no underscores, and the title portion of the suggested validator name is capped at 22 characters.  This makes for a max of 32 characters by the time the '_YYMM' and sometimes '.poem' are added.

Fixes #74 - we already had the translation string for "Select a language for your template", so I've added it as a heading and moved the language selection box down from the corner.

Fixes #43 - the 'prep' function would lowercase the tag names and trim them.  I've replaced calls to 'prep' with calls to 'trim', and removed the now-redundant function.

### DIFF
--- a/application/controllers/private/Validator.php
+++ b/application/controllers/private/Validator.php
@@ -56,7 +56,7 @@ class Validator extends Private_Controller
 		if (empty($this->data['project'])) redirect(base_url() . 'private/validator/select_project');
 
 		$this->data['project']->full_title = create_full_title($this->data['project']);
-		$this->data['project']->suggested_title = url_title(replace_accents($this->data['project']->full_title), '_', true);
+		$this->data['project']->suggested_title = substr(url_title(replace_accents($this->data['project']->full_title), '', true), 0, 21);
 		$this->data['project']->author_full_name = $this->_get_author_by_project($project_id);
 		$this->data['project']->author_last_name = $this->_get_author_by_project($project_id, 'last');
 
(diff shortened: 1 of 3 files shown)


redrun45:
For visual reference, here's where the template language selection was moved for #74:
Screenshot_fix74

redrun45:
Splitting these up into separate PRs.

PR #180, opened 2023-12-30
[WIP] Move template language selection.

### LINKED ISSUE #74: In Project Template Generator, move the box for the different language choices in a more conspicuous spot

redrun45:
It's now the first thing to choose, rather than up in the corner.  Someone already had the 'Select a language for your template' string translated and ready for this change.
Fixes #74 
Sorry for the mess of PRs.


redrun45:
Pending a check on the translations by other admin.
On the technical side, I don't know whether this should be inside a "fieldset" HTML element, and if that's likely to cause other trouble.  I copied it from the other elements below, but this chooser isn't part of the "form" proper.

redrun45:
The positioning on this doesn't look quite right in hindsight.  I'll take another stab at defining CSS classes, and propose again after admins think it looks OK.

PR #182, opened 2023-12-30
[WIP] Keep capitalization on keywords/tags

### LINKED ISSUE #43: In keyword entry on project template, keep capitalization as entered

redrun45:
Fixes #43
The 'prep' function would call 'trim' and 'strtolower' on any given keyword.  I've replaced both calls to 'prep' with just 'trim'.  No more lowercase, and we can remove that function.

### DIFF
--- a/application/libraries/Keywords.php
+++ b/application/libraries/Keywords.php
@@ -106,20 +106,7 @@ class Keywords {
 	 */
 	public function add($keyword)
 	{
-		return $this->ci->keyword_m->insert(array('value' => self::prep($keyword)));
-	}
-
-	/**
-	 * Prepare Keyword
-	 *
-	 * Gets a keyword ready to be saved
-	 *
-	 * @param	string	$keyword
-	 * @return	bool
-	 */
-	public function prep($keyword)
-	{
-		return strtolower(trim($keyword));
+		return $this->ci->keyword_m->insert(array('value' => trim($keyword)));
 	}
 
 	/**
@@ -143,7 +130,7 @@ class Keywords {
 		$keywords = explode(',', $keywords);
 		foreach ($keywords as &$keyword)
 		{
-			$keyword = self::prep($keyword);
+			$keyword = trim($keyword);
 			// Keyword already exists
 			if (($row = $this->ci->db->where('value', $keyword)->get('keywords')->row()))
 			{


redrun45:
I have some technical uncertainty on this one.  I don't see anywhere that uppercase or mixed-case keywords should be an issue, but I don't know that I'd find such trouble if there were.  I've tested a few on my dev instance, and not seen anything, but a closer look at the code would be appreciated.

notartom:
Assuming these are stored in the `librivox_catalog.keywords` table, I don't see keeping capitalization being a technical problem. `keywords`.value is a `TEXT` column, with a UTF8 charset and collation. On the other hands, I can't help but wonder about the user impact of making "Example" and "example" be different keywords, and the confusion that it might entail. Is searching by keywords a thing? Could we end up with a situation in which a book shows up under "Example" and not "example" (asking out of ignorance)?

notartom:
OK, I went digging, and I'm actually against the principle of this.

In [1] we use the keywords to set the `$params['subject']`, which we then sent to IA as the subject piece of metadata at [2]. These keywords/subjects then show up as the Topics on a book's page, ex: https://archive.org/details/a_modern_grandmother_2312.poem_librivox has "librivox, audiobooks, poetry, literature, nostalgia, philosophy, grandmothers". Individual keywords/subjects are clickable, and can be used to search by that subject.

Splitting, for example, "poetry" and "Poetry" into different keywords/subjects seems like an anti-feature to me.

[1] https://github.com/LibriVox/librivox-catalog/blob/master/application/controllers/private/Iarchive_upload.php#L47
[2] https://github.com/LibriVox/librivox-catalog/blob/master/application/libraries/Iarchive_uploader.php#L70

redrun45:
Thanks for taking a closer look.  If this causes trouble, we can do without.

So far as Archive is concerned, it seems that things are case-insensitive (clicking "poetry" returns "Poetry" and "POETRY" results), so I don't _think_ that's a problem. But this is exactly the kind of thing I didn't know to ask!

redrun45:
Re-reading your first comment, it sinks in that we'd be creating new tag/keyword entries in the database for each day this particular code was in production. Fixing #43 without causing that issue may be possible, but not this way. Closing this PR, and we may try something else in future.

PR #183, opened 2023-12-30
[WIP] Confirm before deleting rows in section compiler

### LINKED ISSUE #35: Add confirmation for deleting rows in section compiler
can a confirmation dialog box be added to ask for confirmation when deleting rows in the section compiler (similar to what happens in the Validator)?

Priority: 2

redrun45:
Fixes #35.
Validator uses the 'jquery.alerts' library for its dialog boxes.  I've copied the usage from there:
- Add JS and CSS into the template for Section Compiler
- Copy the confirmation dialog usage into the delete button's 'click' callback


redrun45:
Pending consensus, and I'll post back here when we have the OK.
A screenshot:

Screenshot_fix35

PR #184, opened 2023-12-30
More accurate pagination URLs for Reader, Group, and Author pages

garethsime:
Before this change, the pagination URLs just pointed to `#`, which meant that you couldn't do things like 'Open in new tab', or bookmark a specific page if you were sorting by Release Date.

This commit helps solve those problems on the Reader, Group, and Author pages by:

 * Specifying a better URL for the pagination links (solves for 'Open in new tab')
 * Pushing the `search_order` parameter into the page URL when new pages are fetched from the server (solves for accurate bookmarking)

The changes are more or less the same for each of the three pages, except groups don't have a sort order.

(I haven't fixed it on the Search page, it's got a bit more going on in terms of query params etc.)


application/views/catalog/partials/footer.php
```diff
@@ -45,7 +45,7 @@ function get_search_page_from_url() {
 	var search_page = get_search_page_from_url();
 	set_advanced_form_page(search_page);
 
-	var search_order = 'alpha';
+	var search_order = $('.js-sort-menu').val() || 'alpha';
```
garethsime, line 48: `.js-sort-menu` is the sort order `<select>`. We're already going to some pains to keep it in sync with `search_order`, so I figured it wouldn't be the worst thing if I initialized this variable based on that (if it's available)

application/controllers/catalog/Author.php
```diff
@@ -21,6 +21,8 @@ public function index($author_id)
 		$matches = $this->_get_all_author($author_id);
 		$this->data['matches'] = count($matches);
 
+		$this->data['search_order'] = $this->input->get('search_order');
```
garethsime, line 24: This gets used to mark the correct sort option as selected in `views/catalog/author.php`

redrun45:
I've tested this with the latest master, and it seems to do exactly what it says on the tin.
- Making the search order accessible in this way makes perfect sense to me
- I'm not qualified to review the new function parameter to the core pagination, but I do like how it's documented.

I've kicked the tires a bit, and nothing crashes when I give unexpected values in the URL
- Negative or zero page numbers are reset to 1
- Too-large page numbers show no results, but have controls for returning to page 1
- Unknown values for search_order are reset to 'alpha'

redrun45:
This looks like a good change to me, but I'll be checking with other admins before I say we're ready for it.  So: "pending consensus".  😉

garethsime:
Awesome, thanks 🙂 I'll be away for a few days starting tomorrow, so no rush since I won't be near my computer anyway

notartom:
@redrun45 want me to deploy this on staging to let other admins play around? I still haven't figured out a sane way to make staging multi-user, and allow both of us to "demo" PRs so to speak, so for now we're stuck with just me doing as you direct :)

redrun45:
@notartom that sounds great!  I'm not so much directing as giving a little push, then letting things take the path of least resistance.  :laughing: 

So yes, if this is the next one you'd like to test, let's do it.  I'll show the other admins what's different, ask them to try it out, and get any feedback we can.  It looks like there are some merge conflicts since my last review, but I can _try_ to get those resolved either today, or tomorrow evening, if that would be helpful to you.

garethsime:
@redrun45 I've pushed a merge to fix the conflicts :slightly_smiling_face:

notartom:
Sweet, I've checked our this PR on [staging](https://forum.librivox.org/viewtopic.php?t=69859).

redrun45:
Alright!  I think we'll let things settle after the update tomorrow, and then I'll point folks there to try it out on their own timing.

redrun45:
I updated the timeline on this one in the admin thread: it was a bit much to ask folks to try it out, along with the new changes, without a weekend to do it in.

redrun45:
After two weeks, I think we've tested this as much as there is interest in doing so.  No strange behaviors I could find, this does just what it should.  Admin consensus: :+1:

notartom:
@garethsime looks like this needs a rebase

notartom:
I took the liberty of rebasing this myself, and proposed https://github.com/LibriVox/librivox-catalog/pull/215

notartom:
Done in https://github.com/LibriVox/librivox-catalog/pull/215

garethsime:
Awesome, thanks all!

PR #193, opened 2024-02-12
Expand tests for description HTML formatting

garethsime:
I posted this test as an example when I suggetsed adding ci-phpunit-test, but I had left out a bunch of the actual test cases to keep the description short :slightly_smiling_face:


notartom:
Awww yisss

PR #198, opened 2024-02-24
Remove Google Analytics

notartom:
We've had some users complain about the Ghostery extension flagging
these. As far as we know, they're unused, so remove them.

### DIFF
--- a/application/views/catalog/partials/footer.php
+++ b/application/views/catalog/partials/footer.php
@@ -462,18 +462,3 @@
 
 
 </script>
-
-<script type="text/javascript">
-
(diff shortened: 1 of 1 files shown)


notartom:
@redrun45 so how confident are you that this is fine to remove? I don't know much about Google Analytics, but I imagine there's a management console somewhere that this used to report into?

redrun45:
None of the current admins have access to the Google Analytics account at all, so there's no manual process we'd break.  Archive folks wouldn't have it either, and I don't see anything in our code that might use it programmatically...
I think we can be pretty sure about this one.

Issue #188, opened 2024-02-07
Preserve line/paragraph breaks in description from Template Generator


twinkietoes-on:
A reader fills in the Template Generator, adding paragraphs to the description. These show as paragraph breaks in the database entry, but they're lost at cataloging, both on the catalog page and on the data sent to Archive.org.
![Capture1](https://github.com/LibriVox/librivox-catalog/assets/70523393/c9fdf965-babe-46f2-abef-c90bc959cea6)
![Capture2](https://github.com/LibriVox/librivox-catalog/assets/70523393/de72180a-2328-46d1-a91c-bf1879ff2095)

Current fix is to manually enter `<br />` html to the database field. (Paragraph tags cause other problems in the display.)

twinkietoes-on:
If this involves editing the code that goes to Archive, I'd like to see a couple additional line breaks in there, between the different boilerplate components (top and bottom) and the description, to separate them into paragraphs as well. That's not as critical, but a "might as well, while you're in there" item.

garethsime:
Getting all the newlines to show in the catalog view is a one-line change, but it presents an interesting question around what to do with descriptions that are already using `<br />` tags. For example, there might be entries like this in the management view:

> Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec tincidunt auctor ligula. Donec eu lacinia metus. Vestibulum vel felis vel sapien condimentum ullamcorper eget eget massa. Vestibulum pharetra purus eu erat imperdiet, vitae dignissim massa malesuada. Curabitur sit amet magna quis elit tristique fringilla. &lt;br /&gt;&lt;br /&gt;
>
> Quisque id lorem leo. Sed iaculis nisl lectus, quis congue leo blandit in. Vivamus facilisis velit ut metus fermentum, ut tristique ex sodales. Proin gravida viverra velit, sed finibus dui mollis a. 

Which would then end up being rendered with extra breaks like this:

> Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec tincidunt auctor ligula. Donec eu lacinia metus. Vestibulum vel felis vel sapien condimentum ullamcorper eget eget massa. Vestibulum pharetra purus eu erat imperdiet, vitae dignissim massa malesuada. Curabitur sit amet magna quis elit tristique fringilla.
> &nbsp;
> &nbsp;
>
[...]

PR #202, opened 2024-03-09
Use spacing helper in Archive uploads, and fix display of older items in the catalog

redrun45:
Got admin consensus for this one:  :+1: 

The regex version will collapse any br that is immediately followed by a \n, a \r, or both, into just a single newline.  That fixes display of old cast lists and such on our catalog pages.  Added a few more test cases to address this, and possible capitalization differences.

Then we use the helper on Archive uploads, so both are consistent.  Also add an extra space before and after the item description, as these are manually added by MCs right now.


redrun45:
Oh, right: per discussion in #188, I've tested this pretty thoroughly on my dev machine, including by archiving some dummy projects.  I think this one is safe to go live.

PR #203, opened 2024-03-09
Add a link to Archive from Advanced Search page

redrun45:
Another one with admin consensus: :+1: 

I tried combinations of things, but it doesn't look bad as just an h5.  And there's no business logic to get wrong.  This can go live at your convenience.  :smiley:

### DIFF
--- a/application/views/catalog/partials/advanced_search.php
+++ b/application/views/catalog/partials/advanced_search.php
@@ -101,6 +101,10 @@
 	         </div>
 	       </div>
 
+               <h5>Still can't find what you're looking for? Try searching <a rel="noopener" target="_blank" href="https://archive.org/details/librivoxaudio">Internet Archive's LibriVox Audiobook Collection</a>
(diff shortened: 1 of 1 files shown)


application/views/catalog/partials/advanced_search.php
```diff
@@ -101,6 +101,10 @@
 	         </div>
 	       </div>
 
+               <h5>Still can't find what you're looking for? Try searching <a rel="noopener" target="_blank" href="https://archive.org/details/librivoxaudio">Internet Archive's LibriVox Audiobook Collection</a>
```
notartom, line 104: I tend to dislike messing with the defaults for opening links, as it breaks expectations and removes control from the user (middle click to open in new tab only if I chose to), is there a reason you went with _blank? As far as I can tell, this isn't consistent with the other links on the advanced search page either.

redrun45: Good point.  Our 'Internet Archive Page' links from the individual works' pages should probably be the model here.

PR #209, opened 2024-03-16
Normalize RSS spacing

### LINKED ISSUE #188: Preserve line/paragraph breaks in description from Template Generator

redrun45:
The per-title RSS feed (views/rss/rss.php) has been stripping all HTML tags, including spacing.  The latest releases feed has not.  If we're going to strip tags, it should be on the input side instead.

This PR makes use of Gareth's helper, and (I think) fixes #188 for good.

I'd like to re-factor RSS sometime, add in GUIDs to the per-title feed, along with a pubDate... but not today. :neutral_face:

### DIFF
--- a/application/controllers/rss/Rss.php
+++ b/application/controllers/rss/Rss.php
@@ -42,6 +42,9 @@ class Rss extends Catalog_controller
 
 		$this->_build_sections();
 
+		$this->load->helper('description_html_render_helper');
+		$this->data['project']->description = _normalize_and_deduplicate_newlines_in_html($this->data['project']->description);
+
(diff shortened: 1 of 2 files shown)


redrun45:
Forgot to say for this one: we've got the OK, or at least no objections.  :+1:

PR #217, opened 2024-04-06
Fix API giving uploader URLs instead of Archive

### LINKED ISSUE #211: API gives temporary section URLs
**Expected behavior:**
When calling the LibriVox API with 'extended=1', it is supposed to list the URL of each section's MP3 file, available for download.

**Actual behavior:**
Currently, it gives the URL we use for projects that are _in progress_, rather than the final Archive.org download URL.  These will look like `https://librivox.org/uploads/<mc>/<file>.mp3`, instead of `https://www.archive.org/download/<project>/<file>.mp3`.

This is not the right URL because:
* It puts load on our particular server, rather than the distributed infrastructure for finished files.  We ask that folks _only_ download from projects in progress if they are signed up to work on those particular files (e.g., as Proof Listener of a given project).
* At some point after the project is completed and cataloged, these files will disappear from our server, and all of these URLs will return 404 errors.

**Proposal:**
Unless there's an important use-case I'm missing, I think everyone will be happier if this returned the Archive.org URLs for sections of _finished projects only_, exactly the way `url_zip_file` works now.
We could even give both the URLs for both the original-quality 128kb recordings, and the 64kb versions derived by Archive.org.

Posting this here before I open a Pull Request, so that anybody who depends on this API has a chance to see it first.

redrun45:
Fixes #211.  Calling `/api/feed/audiobooks` with `extended=1` will now give the Arhive.org URLs for sections in completed projects, and empty strings for projects in progress.
From the database, we're substituting `mp3_64_url` for the previously used `listen_url`.  I opted to leave the API name the same, in case someone actually has been using it successfully.

Side note: I've added several `COALESCE()` calls, to get rid of some warnings.  They might not do anything but clutter the logs in prod, but on localdev they get rendered as HTML, making the API more difficult to test.


notartom:
I'm not convinced the use of COALESCE is correct here...

So, for context, the errors that you speak of are from passing a `$value` of NULL to https://github.com/LibriVox/librivox-catalog/blob/master/application/libraries/Format.php#L142. So the errors only show up when looking at the results in XML format - passing `&format=json` in the API URL makes them go away.

I feel like the better fix is to add an `if ($value) { ... } else { $value = '' };` - it addresses the error closest to its source, and makes it easier to understand why it's being done. Using COALESCE in the SQL query is "spooky action at a distance".

Also, at least in my tests on staging, using COALESCE doesn't make all the errors disappear.

And finally, because COALESCE returns the first non-NULL element in the list, I _think_ that the values that these COALESCE statements return aren't even guaranteed to come from the same section. All you need is a different number of NULLs in font of the first non-NULL, and suddenly you're pulling the URL from the 3rd section, the language fromthe 42nd, etc.

redrun45:
Alright! I'll give this one another stab.  I saw COALESCE apparently used for this purpose, and figured we should probably handle any other 'nullable' values the same way.

Marking as [WIP] until this is addressed.  Ran out of coding time for the evening.  😉

application/libraries/Format.php
```diff
@@ -139,7 +139,13 @@ public function to_xml($data = null, $structure = null, $basenode = 'xml')
 			else
 			{
 				// add single node.
-				$value = htmlspecialchars(html_entity_decode($value, ENT_QUOTES, 'UTF-8'), ENT_QUOTES, "UTF-8");
+				if (isset($value)) {
```
notartom, line 142:
```
artom@librivox-staging:~$ php -a
Interactive shell

php > isset(NULL);
PHP Fatal error:  Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in php shell code on line 1
php >
```

notartom (changes requested):

notartom:
I was convinced `isset()` on a NULL value wouldn't work, and it _doesn't_, but `isset()` on a _variable_ that is NULL works as expected. Gdi PHP:

```
php > isset(NULL);
PHP Fatal error:  Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in php shell code on line 1
php > $foo = NULL;
php > if (isset($foo)) { print("set"); }
php > if (! isset($foo)) { print("set"); }
set
php > 
```

notartom:
Yeah - so anyways, apologies for the hasty review, and I'm not sure what happened to your branch, but please re-open and repropose :)

PR #220, opened 2024-04-11
Fix Archive 'Read by' list not matching catalog

### LINKED ISSUE #201: Archive description lists readers once per section, not once per project
Per discussion [here](https://github.com/LibriVox/librivox-catalog/issues/82#issuecomment-1002216115): since we moved away from 'select DISTINCT', which was slowing things down, we now get duplicates of reader names on the Archive pages.

Someone reading a book solo would have their name repeated for each section, like so:
Read in English by ThisReader; ThisReader; ThisReader; ThisReader; ThisReader; ThisReader; 

Looking at the database tables, I think we can 'select from project_readers', and get only one listing per reader.  I'll be testing that on my dev machine, as and when I work out how best to do that.  :laughing: 

@twinkietoes-on - for solo projects, showing the reader name only once is obviously an improvement.
But for group projects, do we _want_ to list all the readers on Archive?  It might make sense to name both readers for a duet, where a project with a dozen readers probably ought to say 'read by LibriVox Volunteers'.  Is there a particular number we should cut off at - 1, 2, ...5?  :thinking:

redrun45:
Fixes #201.  Per MC discussion, we want Archive pages to match the existing 'Read by' label on our catalog pages:
* All types of group projects show 'LibriVox Volunteers', regardless of reader count.
* Solo projects show the first reader result returned from the database, or 'Solo' if there are none.

Along with this change, I've refactored `get_project_readers` and `create_project_reader_list` into `get_solo_reader`:
* `get_project_readers` was called only from catalog pages, but only for solo projects, and only to get the first reader result.
* `create_project_reader_list` is no longer needed.  It was called only from Iarchive_upload, which should also only need to call it for solo projects.

If you think it would be better to have the group vs solo check be in the model rather than the controllers, we could do that at the cost of a second (but quick) database query.  Open to that, or other suggestions!

### DIFF
--- a/application/controllers/catalog/Page.php
+++ b/application/controllers/catalog/Page.php
@@ -131,9 +131,9 @@ class Page extends Catalog_controller
 	{
 		if (strtolower($project->project_type) == 'group') return 'LibriVox Volunteers';
 
-		$readers = $this->project_model->get_project_readers($project->id);
+		$readerstub = $this->project_model->get_solo_reader($project->id);
 
-		if ($readers) return '<a href="' . base_url() . 'reader/' . $readers[0]['reader_id'] . '">' . $readers[0]['display_name'] . '</a>';
+		if (isset($readerstub)) return '<a href="' . base_url() . 'reader/' . $readerstub['reader_id'] . '">' . $readerstub['display_name'] . '</a>';
 
 		return 'Solo';
 	}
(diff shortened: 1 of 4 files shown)


application/controllers/catalog/Page.php
```diff
@@ -131,9 +131,9 @@ function _read_by($project)
 	{
 		if (strtolower($project->project_type) == 'group') return 'LibriVox Volunteers';
 
-		$readers = $this->project_model->get_project_readers($project->id);
+		$readerstub = $this->project_model->get_solo_reader($project->id);
 
-		if ($readers) return '<a href="' . base_url() . 'reader/' . $readers[0]['reader_id'] . '">' . $readers[0]['display_name'] . '</a>';
+		if (isset($readerstub)) return '<a href="' . base_url() . 'reader/' . $readerstub['reader_id'] . '">' . $readerstub['display_name'] . '</a>';
```
notartom, line 136: The fact that we return inline HTML from the controller is janky, but you didn't write that, and it's out of scope in this PR to fix.

Issue #197, opened 2024-02-23
Be able to narrow down the author choice in Template Generator


twinkietoes-on:
https://librivox.org/add_project

When typing in a common surname for the author, or one that is a partial match to a lot of other names, it returns a long list. Example: typing in "Smith" gives you "Grossmith" and every other surname that contains "Smith" somewhere in it, and the list is getting unwieldy. There's no way to narrow it down (for example, you can't type the first/given name to narrow the list).

I don't know the best solution, but we do need a way to narrow down this list for people to find the author to select; otherwise, they'll give up and make a duplicate author.

peterjdann:
Seems to me the behaviour we see here would be better if:

- In a case where the user has specified only an author surname, the results were sorted by first name. This would mean that "Andrew Smith" would be listed in the dropdown before "Austin Smith" (I made up these examples)
- In a case where the user has specified only an author first name, the results were sorted by author surname. This would mean that "Mary Bolton" was listed before "Mary Boulton" (again, I made this up)

This change would be pretty easy to implement, and while it wouldn't completely solve the OP's problem, it might go some way to allleviating the pain.

peterjdann:
Sorry it's taken me a few days to get to look at this properly. I think the way you've tackled this is far better than my own version. I don't see any value in presenting both options at all.

What is the best way for us to proceed now to get your change implemented? I'd be perfectly happy to have my own pull request aborted/cancelled/deleted (not quite sure how this works) if the easiest way to get this done would be for you to raise a fresh pull request that implements the change whose effects you've illustrated above.

peterjdann:
> Honestly, having two options is good in any case! If you'd like to cancel #232 for now, I have no objections, but we'll at least still have the screenshots. I'll see what other admins think about them, and then I can put up either version (with added pseudonym fix) for technical review. 👍

I've closed #232. I think you solution is much better!

Issue #42, opened 2021-01-13
Author dropdowns should be able to cross-reference pseudonyms from the database


twinkietoes-on:
In Project Generator and the project screen. 

Projects with authors using a name from the pseudonym list cannot access the correct name because they aren't in the pick list. The user often creates a new author entry because they think the name is not in the system. The list should include the pseudonyms.

Priority: 2

PR #223, opened 2024-05-01
Expand author autocomplete

redrun45:
Status updated in a later comment.

----
**Status of this PR:**
* Code: is in need of some further tests before going live.
* "Do we want it?": Precisely the reason for the early PR.  I'd like folks to try it out on staging, when you have the time to put it there.

Until now, autocomplete for filling in author names has been confined to only searching by `last_name` of the "real" author entry in the database.  Pseudonyms are not included in the search list (#42), and for common names like "smith", we have too many results for folks to find the right one (#197)... resulting in duplicates as authors are re-added.  This PR addresses both.

I know this may be one of those 'too clever for your own good' situations, so please do let me know what you think, when you get the time.  I've rewritten this series of commits, which should hopefully make each one easier to follow than the larger chunk all at once.

### DIFF
--- a/application/controllers/public/Public_ajax.php
+++ b/application/controllers/public/Public_ajax.php
@@ -12,8 +12,10 @@ class Public_ajax extends Public_Controller
 	{
 		$term = $this->input->post('term');
 		$search_field = $this->input->post('search_field');
+		$filter_term = $this->input->post('filter_term');
+		$filter_field = $this->input->post('filter_field');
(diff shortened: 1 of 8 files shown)


notartom:
Deployed on staging.

notartom:
I'm not an expert in any of that code (as mentioned previously, the original developer has long since moved on, I'm just the sysadmin), but nothing jumps out as being insane ;) I think we can merge and deploy.

PR #225, opened 2024-05-16
Reveal keywords on project's catalogue page

peterjdann:
The effect of the changes proposed here would be:

- to display a list of keywords associated with a project on that project's catalog page, together with statistics indicating to the viewer how many other projects share each keyword, and 
- to allow the user, by clicking a hyperlinked keyword, immediately to list other projects associated with that keyword. 

In compiling this proposal, I have looked for opportunities to develop any relevant unit tests, but have not seen any.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -2,5 +2,5 @@
 
 $config['migration_enabled'] = TRUE;
 $config['migration_type'] = 'sequential';
-$config['migration_version'] = 3;
+$config['migration_version'] = 4;
 $config['migration_path'] = APPPATH . 'migrations/';
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -2,5 +2,5 @@
 
 $config['migration_enabled'] = TRUE;
 $config['migration_type'] = 'sequential';
-$config['migration_version'] = 2;
+$config['migration_version'] = 3;
 $config['migration_path'] = APPPATH . 'migrations/';
--- a/application/controllers/catalog/Page.php
+++ b/application/controllers/catalog/Page.php
@@ -61,6 +61,20 @@ class Page extends Catalog_controller
 		$this->data['volunteers']->bc = $this->user_model->get($this->data['project']->person_bc_id);
 		$this->data['volunteers']->mc = $this->user_model->get($this->data['project']->person_mc_id);
 		$this->data['volunteers']->pl = $this->user_model->get($this->data['project']->person_pl_id);
+	
+		
+		// **** KEYWORDS ****//
+		$MAX_KEYWORDS_TO_DISPLAY_INITIALLY = 10;
+		$this->data['keywords_array'] = $this->project_model->get_keywords_and_statistics_by_project($this->data['project']->id);
+		$this->data['project']->has_long_keywords_list = $this->_has_long_keywords_list($this->data['project']->id, $MAX_KEYWORDS_TO_DISPLAY_INITIALLY);
+		$this->data['project']->formatted_keywords_string_primary = $this->_formatted_keywords_string($this->data['project']->id, 
+													$this->data['project']->has_long_keywords_list, 
+													$MAX_KEYWORDS_TO_DISPLAY_INITIALLY, 
+													"primary_list");
+		$this->data['project']->formatted_keywords_string_secondary = $this->_formatted_keywords_string($this->data['project']->id, 
+													$this->data['project']->has_long_keywords_list, 
+													$MAX_KEYWORDS_TO_DISPLAY_INITIALLY, 
+													"secondary_list");
 
 		// **** MISC ****//
 		$this->data['project']->project_type = (trim($this->data['project']->project_type) == 'solo') ? 'solo' : 'group'; //keep on top - other values dependent on solo/group
@@ -170,7 +184,123 @@ class Page extends Catalog_controller
 
 		return implode(', ', $genres);
 	}
-
+	
+	function _formatted_keywords_string($project_id, $project_has_long_keywords_list, 
+						$max_keywords_to_display_initially, $list_variant)
+	{
+	/** 	The catalog page will contain one, or possibly two lists of keywords.
+		It will always contain a primary list.
+		If the project has more keywords than can conveniently be fitted in the primary list:
+		(a) We truncate this primary list
+		(b) We also output to the page a secondary list, iniitially hidden, that includes all the keywords for the project
+		(c) We add to the primary list a "Show full list" link which, if clicked, hides the primary list 
+		and shows the secondary list
+		*/
+		
+		$return_value = "";
+		
+		switch ($list_variant) {
+  			case "primary_list":
+  				if ($project_has_long_keywords_list)
+  				{
+  					$return_value =	$this->_formatted_truncated_keywords_string($project_id, $max_keywords_to_display_initially);
+  				} 
+  				else
+  				{
+  					$return_value =	$this->_formatted_full_keywords_string($project_id); 
+  				}
+    				break;
+  			case "secondary_list";
+    				if ($project_has_long_keywords_list)
+  				{
+  					$return_value =	$this->_formatted_full_keywords_string($project_id);
+  				} 
+    				break;
+  			default:
+    				$return_value = "";
+		}
+		return $return_value;
+	}
+	
+	
+	function _formatted_full_keywords_string($project_id) 
+	{
+		$return_value = '';
+		$keywords_array = $this->data['keywords_array'];
+		if (!empty($keywords_array))
+		{
+			foreach ($keywords_array as $key => $row)
+			{
+			// Add hyperlink only if at least two projects use this keyword
+				if ((int)$row['keyword_count'] > 1) 
+				{
+					$return_value .= '<a href="' . base_url() . 'keywords/' . $row['id'] . '">';
+				} 
+				$return_value .= $row['value'];
+				if ((int)$row['keyword_count'] > 1)
+				{
+					$return_value .= '</a>';
+				} 
+				$return_value .= ' (' . $row['keyword_count'] . '), ';
+			}
+		} 
+		else 
+		{
+			return $return_value;
+		}
+		// trim off final comma and trailing space 
+		$return_value = substr($return_value, 0, -2);
+		return $return_value;	
+	}
+	
+	function _formatted_truncated_keywords_string($project_id, $max_keywords_to_display_initially) 
+	{
+		$return_value = '';
+		$keywords_array = $this->data['keywords_array'];
+		if (!empty($keywords_array))
+		{
+			$i = 0;
+			foreach ($keywords_array as $key => $row)
+			{
+				
+			// Add hyperlink only if at least two projects use this keyword
+				if ((int)$row['keyword_count'] > 1) 
+				{
+					$return_value .= '<a href="' . base_url() . 'keywords/' . $row['id'] . '">';
+				} 
+				$return_value .= $row['value'];
+				if ((int)$row['keyword_count'] > 1)
+				{
+					$return_value .= '</a>';
+				} 
+				$return_value .= ' (' . $row['keyword_count'] . '), ';
+				$i++;
+				if ($i == $max_keywords_to_display_initially)
+				{
+					break;
+				}
+			}
+		} 
+		else 
+		{
+			return $return_value;
+		}
+		// trim off final comma and trailing space 
+		$return_value = substr($return_value, 0, -2);
+		
+		// Add "Show full list" link
+		$return_value .= " ... <a href='#' class='truncated-keywords js-truncated-keywords'>[Show full list]</a>";
+		
+		return $return_value;	
+	}
+	
+	function _has_long_keywords_list($project_id, $max_keywords_to_display_initially)
+	{
+		$keywords_number = $this->project_model->get_keywords_count_by_project($project_id)->keywords_number;
+		return $keywords_number > $max_keywords_to_display_initially;
+	}
+	
+		
 	function _language($language_id)
 	{
 		$this->load->model('language_model');
--- a/application/controllers/cron/Keywords_stats.php
+++ b/application/controllers/cron/Keywords_stats.php
@@ -1,0 +1,32 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Keywords_stats extends CI_Controller
+{
+	public function update_keywords_stats()
+	{
+		// Some keywords are not in use at all.
+		// Initialise count field to zero before tallying counts.
+		
+		$sql = '
+			UPDATE keywords
+			SET count = 0';
+		$this->db->query($sql);
+		
+		$sql = '
+			UPDATE keywords
+			JOIN
+			(SELECT pk.keyword_id as keyword_id, COUNT(pk.project_id) AS count
+        		FROM project_keywords pk
+        		GROUP BY 1
+			) as sub
+			ON keywords.id = sub.keyword_id
+			SET keywords.count = sub.count ';
+		$this->db->query($sql);
+
+	}
+}
+
+
+/* End of file Keyword_stats.php */
+/* Location: ./application/controllers/cron/Keyword_stats.php */
+
--- a/application/migrations/004_drop_redundant_keywords_from_projects.php
+++ b/application/migrations/004_drop_redundant_keywords_from_projects.php
@@ -1,0 +1,52 @@
+<?php
+
+defined('BASEPATH') OR exit('No direct script access allowed');
+
+class Migration_Drop_redundant_keywords_from_projects extends CI_Migration {
+
+        public function up()
+        {
+    		$sql = '
+		DELETE pk2.*
+		FROM project_keywords pk2
+		WHERE keyword_id IN
+		(
+			SELECT id 
+			FROM keywords k 
+			WHERE k.value in (
+			"librivox", 
+			"ibrivox",
+			"llibrivox",
+			"librixox",
+			"librivox;",
+			"libriox",
+			"libirivox",
+			"librivos",
+			"audio",
+			"audiobook",
+			"audiobooks",
+			"audio books",
+			"audio book",
+			"audioboek",
+			"audiobuch",
+			"audiobooklibrivox",
+			"audibook",
+			"audioook",
+			"autiobook",
+			"audiobook:",
+			"audioboo",
+			"audiolibro",
+			"audio libro",
+			"audiolibros",
+			"audiolivre",
+			"audiolivro"
+			)
+		)';
+    		$this->db->query($sql);
+        }
+
+        public function down()
+        {
+                ;
+        }
+}
--- a/application/models/Project_keyword_model.php
+++ b/application/models/Project_keyword_model.php
@@ -2,7 +2,27 @@
 
 class Project_keyword_model extends MY_Model {
 
-    
+   public function set_keywords_statistics_by_project($project_id)
+    {        
+        $sql = '
+ 	UPDATE keywords k
+	JOIN
+		(SELECT pk.keyword_id as keyword_id, COUNT(pk.project_id) AS count
+        	FROM project_keywords pk
+		GROUP BY 1
+		) as sub
+	ON k.id = sub.keyword_id
+	SET k.count = sub.count 
+	WHERE k.id IN
+		(SELECT keyword_id 
+        	FROM project_keywords
+        	WHERE project_id = ?)';
+
+
+        $query = $this->db->query($sql, array($project_id));
+    }
+
+
 }
 
 /* End of file project_keyword.php */
--- a/application/models/Project_keyword_model.php
+++ b/application/models/Project_keyword_model.php
@@ -3,7 +3,6 @@
 class Project_keyword_model extends MY_Model {
 
     
-
 }
 
 /* End of file project_keyword.php */
--- a/application/models/Project_model.php
+++ b/application/models/Project_model.php
@@ -363,6 +363,85 @@ class Project_model extends MY_Model
         return $query->result();
 
     }
+    
+    
+    public function get_projects_by_keywords_id($params)
+    {
+	// If calls to this function do not pass a project_type_description parameter,
+	// provide a sensible default
+		
+	$keyword_id = $params['keywords_id']; 
+        $offset = 0 + $params['offset'];
+        $limit = 0 + $params['limit'];      
+        
+        $sql = 'select p.id, p.title_prefix, p.title, COALESCE(p.date_catalog, "2001-01-01" ) as safe_release_date, 
+        p.url_librivox, p.status, p.coverart_thumbnail, 	        p.zip_url, p.zip_size, l.two_letter_code, l.language, p.url_forum, p.project_type
+	FROM keywords k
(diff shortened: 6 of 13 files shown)


peterjdann:
I am puzzled by the error message: The migration class \"Migration_Add_keywords_instances_field\" is missing an \"up\" method."

I have the following settings in catalog/application/config/migration.php:

$config['migration_enabled'] = TRUE;
$config['migration_type'] = 'timestamp';
$config['migration_version'] = 20240516002600;
$config['migration_path'] = APPPATH . 'migrations/';

The following are the contents of catalog/application/migrations/20240516002600_add_keywords_instances_field.php

<?php

defined('BASEPATH') OR exit('No direct script access allowed');

class Migration_Add_keywords_instances_field extends CI_Migration {

        public function up()
        {
		$fields = array(
        		'instances' => array('type' => 'INT',
			'default' => '1',
			'after' => 'value')
		);
		$this->dbforge->add_column('keywords', $fields);
        }

        public function down()
        {
                $this->dbforge->drop_column('keywords', 'instances');
        }
}

peterjdann:
Thank you very much for this helpful initial feedback.

I did not know it was possible to “squash” git commits until now, but have just looked up some info on that, and see it is possible. Sorry you’ve having to deal with such a git (and software review) newbie here. Will wait for advice from Artom on that, as you suggest.

I’m away from my development environment for the next four days. I will look more carefully at your other suggestions as soon as I am back. Your explanations all seem very clear to me.

Before I start making any changes, I would like to wait on whatever initial advice Artom may have on how to proceed from here.

Can I assume that where I make changes to address the issues you have raised, I should, as a matter of principle, ’squash’ the commit for each further change with the previous commit affecting the same area of code?

Peter Dann

peterjdann:
I have had a go at addressing what I think are the major concerns redrun has expressed so far (though I'm afraid I don't understand redrun's point about the cron job endpoint, and have not addressed that). I have now had a (reasonably brutal!) first person introduction to rewriting a git commit history — the need for such, the means at one's disposal, and the many (many!) pitfalls lying in wait for the beginner. I hope I've made some useful progress in this area.

peterjdann:
Thanks for that explanation of the cron job issue (which should have been staring me in the face). I've fixed that now, and have pushed that change and all "subsequent" changes here again.

peterjdann:
> Well, I've finally sat down to configure my dev box to catalog mock projects again. I need to do a PR to add this to the Ansible scripts sometime. 🤔
> 
> That was the last test I knew we _couldn't_ do on the staging server. If you'd like to implement the changes I previously mentioned, I won't hold you up any longer, and it can go to staging.
> 
> For convenience, here are quick summaries of those three things:
> 
> * Technical - the `.gitignore` file being removed from the repo
> * Technical - `Project_keyword_model.php` line 19 concatenating the project id, instead of using a '?'. (Line 22 sets up the DB library so that it will escape and substitute the id in place of the '?', which is the preferred method.)
> * Usability suggestion - in `views/catalog/keywords.php`, add a header to display the particular keyword we're browsing.
> 
> Once at least the first two are addressed, we'll ask if Artom can load it up on Staging for more people to play with. Perhaps along with #231, for the full experience! 😉
> 
> Then on a more personal note, I know that text isn't my best medium for building empathy. Code reviews, in particular, seem very "transactional" to me, as they're all focused on business. They'd be harder to read, if they weren't! But that can make them, if not _antagonistic_, at least a little adversarial, which I know _I_ find frustrating from the relationship side.
> 
> I want to say: **Thank you!** Thank you for caring in the first place to suggest alleviations for the pain-points you see. Thank you for being respectful as you made those suggestions, rather than expecting somebody to "_just_" do _X_.
> 
> In our forum conversation, I was most explicit in saying that I _would not ask_ someone to work on our code, for reasons that extend beyond the messy code itself. But I'm glad you did. _Thank you again_ for volunteering - and learning, and returning - to work on this anyway.
> 
> To make this happen, we needed you. With the amount of new code this took, and the number of other bugs and quirks that have **bothered** volunteers for years, I don't know if I'd have ever gotten to such a big new feature. Between you and Gareth (and hopefully me), we've made and are making these changes in the past few months, that folks will **enjoy** for years, even if nobody specifically asked for them to happen right now.
> 
> You and I are excited to work on the code, as and when we can. Please be patient with me, as I attempt to save other volunteers (who did not sign up to do code review or QA testing) as much work, worry and stress as possible. Significant updates come with all of those things, for more than just us folks here on GitHub. Even a small update can have unexpected quirks, and we don't want to pressure people to find them _before they have to live with them_. (See also: #233 😓 ) Thank you for your patience so far, and I hope we'll all get to see this live fairly soon!

I have now attempted to address all three of the technical issues you have (very conveniently - thank you!) summarised here and have pushed the changes through.
--------
On the personal side, I doubt if you can have any idea how glad and thankful I was to receive your very kind, generous and timely message. You have been the soul of discretion, and the soul of helpfulness, through this whole process, for ever acting in what I have long since come to recognise as the true Librivox spirit. If I can live up to that ethos half as well as you do, I will be very glad indeed (😃) - and probably a bit surprised, too!

peterjdann:
> Excellent, I'm glad my nit-picking hasn't driven you away just yet! So far as I'm concerned, the LibriVox spirit is: "volunteers, helping volunteers." I know you've got the same spirit in mind, and I just hope for a reminder whenever I get lost in the weeds myself. 😄
> 
> Back to the business side, those changes address everything I know we need for this PR. You mentioned you might make changes to the other one, but I think we can call this one "Ready for Staging"!
> 
> * In terms of what it _does_: Looks good to me, but wiser heads than mine should do some further testing.
> * In terms of code: Similarly, if Gareth, Artom or anyone has further suggestions... 😉

I have already made those "further changes" to #231 . Those changes, if acceptable for staging yet, might complete the changes here nicely.

application/controllers/cron/Keywords_stats.php
```diff
@@ -0,0 +1,27 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Keywords_stats extends CI_Controller
+{
+	public function update_keywords_stats()
+	{
+		$sql = '
```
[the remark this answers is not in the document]
peterjdann: Excellent bug fix! Thank you! I've pushed that change through now.
(By my count we currently have just over 2900 unused keywords in our database, including such gems as "i can send more later", "I will do these before cataloging" and "i will provide these before i'm finished with the project".
Maybe this same job should be doing a cull of unused keywords too? (I'm not going to code that now - just floating it as a suggestion for further discussion.)

peterjdann:
It's possible that when administrators see this capability in action, they'll be quietly (or not so quietly) horrified at some of the keywords they see are in actual use. It occurs to me that one way to alleviate this horror ever so slightly might be to provide a function that made it easy for administrators quickly to edit those keywords. While it's not necessarily ideal for this purpose, the "add_catalog_item" page does support this functionality, and it's a relatively simple exercise to add to the project catalog page a link, visible only to logged in administrators, that would take them straight to this page for the project concerned. I've mocked that up on my local development system as a proof-of-concept. Hopefully the attached document illustrates properly what I mean.
[Support easy editing of keywords for existing projects.pdf](https://github.com/user-attachments/files/15983715/Support.easy.editing.of.keywords.for.existing.projects.pdf)

peterjdann:
> This link looks like a nice convenience feature in general! It's much more useful for other things than for keywords, though. 
> ... 
> Perhaps this new link should be its own thing, over in the side-bar. And though I'm sorry to add more for Artom to sort through, I think a separate PR would be appropriate.

Good idea - but sorry Artom! Have put this forward now as #234

peterjdann:
How do we want to handle a situation where an audiobook has a very long list of keywords associated with it?

One approach would be to truncate the list of keywords we display, showing, say, only the most widely used keywords.

A disadvantage of this approach is that it might bewilder a user who has done an Advanced search for a certain keywords term, the search throws up BOOK X, but BOOK X (which, behind the scenes, turns out to have a long list of keywords) apparently is not associated with this keywords term, to judge by what appears on its Catalog page.

This could be confusing.

The change I have just attempted to add here has the effect of truncating a long list of keywords (for example, more than 10) on the catalog page for audiobooks that have such long lists, of keywords but adding a "Show full list" link which, if clicked, causes the full list to be revealed. For audiobooks whose list of keywords is within the prescribed limit (here coded as 10) the behaviour is the same as it was previously.

Three screenshots follow.

This audiobook has many keywords defined for it. Accordingly, we truncate the list.

Truncated list

When the user clicks "Show full list", they see the whole list of keywords.

Full list revealed

For a book with a relatively brief list of keywords, the behaviour is as previously coded in this PR. There is no "Show full list" link.

List requiring no truncation

peterjdann:
Have added a migration action that drops "librivox", "audiobook" and other terms that appear to be intended to have these meanings from projects. If anyone has any tips on how to test a CodeIgniter 3 migration in our particular setup before conducting one for real, I'd like to hear about it. I'd also like to know if dbforge might, in fact, allow one to frame a delete query involving a subquery. Because I wasn't confident it would, I tried a workaround here.

peterjdann:
> > If anyone has any tips on how to test a CodeIgniter 3 migration in our particular routing setup before conducting one for real, I'd like to hear about it.
> 
> If you mean testing it on your dev box before sending it in as a PR, I use this line (shamelessly copied from Artom's ansible playbook): `php public_html/index.php migrate migrate` If a migration fails (yours worked just fine!), you can re-run it without entirely resetting the database, by altering `migrations.version` to an earlier date.
> 
> > I'd also like to know if dbforge (the orthodox tool for writing migration actions) might, in fact, allow one to frame a delete query involving a subquery. Because I wasn't confident it would, I tried a workaround here, loading and invoking a model function.
> 
> Actually, I don't think we have to use dbforge at all! I moved your SQL query from Project_keyword_model straight into the migration file, and it worked just fine. Looks like dbforge is loaded in as a convenience, but we still have the broader range of options, including `$this->db->query()`. 😁

Thank you! That's very helpful. I did a quick (obviously incomplete) dive into the CI source code and wasn't able to convince myself that $this->db would actually be available, and thinking I had no easy way to test (short of pushing the change here) I opted for an alternative approach. But doing the database cull entirely inside the migration action seems cleaner to me, so I've now followed that approach, and have removed the use-once-only method from the project_keyword model.

notartom:
Deployed on staging as requested!

application/config/migration.php
```diff
@@ -1,6 +1,6 @@
 <?php defined('BASEPATH') OR exit('No direct script access allowed');
 
 $config['migration_enabled'] = TRUE;
-$config['migration_type'] = 'sequential';
-$config['migration_version'] = 0;
+$config['migration_type'] = 'timestamp';
+$config['migration_version'] = 20240516002600;
```
notartom: Let's stick to `sequential`, and since https://github.com/LibriVox/librivox-catalog/pull/239 has merged, this should now be `2` :)

system/libraries/Migration.php
notartom, line 1: Once you rebase on top of current master, the changes to this file should go away, since they were incorporated into https://github.com/LibriVox/librivox-catalog/pull/239.

application/migrations/20240516002600_add_keywords_instances_field.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+defined('BASEPATH') OR exit('No direct script access allowed');
+
+class Migration_Add_keywords_instances_field extends CI_Migration {
+
+        public function up()
+        {
+		$fields = array(
+        		'instances' => array('type' => 'INT',
```
notartom, line 10: This is a count, right? Or put differently, how many instances of this keyword is applied to projects? I hate to bikeshed on things like this, but because this will be in our database schema, getting the name right the first time around is important. I would strongly prefer we use `count` rather than `instances`. There's precedent as well: we have a `section_status_count` table.

application/config/migration.php
```diff
@@ -2,5 +2,5 @@
 
 $config['migration_enabled'] = TRUE;
 $config['migration_type'] = 'timestamp';
-$config['migration_version'] = 20240516002600;
+$config['migration_version'] = 20240705003100;
```
notartom, line 5: Going with the comment on the previous commit, this should be `3` :)

application/models/Project_keyword_model.php
```diff
@@ -3,7 +3,6 @@
 class Project_keyword_model extends MY_Model {
 
     
-
```
notartom, line 6: Random unnecessary white space change?

application/migrations/20240705003100_drop_redundant_keywords_from_projects.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+defined('BASEPATH') OR exit('No direct script access allowed');
+
+class Migration_Drop_redundant_keywords_from_projects extends CI_Migration {
+
+        public function up()
+        {
+    		$sql = '
+		DELETE pk2.*
```
notartom, line 10: I initially had a comment about why we're not removing the keywords themselves, only the links to the projects, but in the following commit there's a discussion around keeping all the keywords, even the crazy ones, and I got convinced. This is fine :)

application/controllers/cron/Keywords_stats.php
```diff
@@ -0,0 +1,35 @@
...
+class Keywords_stats extends CI_Controller
+{
+	public function update_keywords_stats()
+	{
+		// Some keywords are not in use at all.
+		// Initialise instances field to zero before tallying instances.
+		
+		$sql = '
+			UPDATE keywords
+			SET instances = 0';
```
notartom, line 12: This would be `count`, assuming we go with my earlier suggestion.

application/controllers/cron/Keywords_stats.php
```diff
@@ -0,0 +1,35 @@
...
+			) as sub
+			ON keywords.id = sub.keyword_id
+			SET keywords.instances = sub.count ';
+		$this->db->query($sql);
+
+	}
+}
+
+/*
+ALTER TABLE `keywords` ADD `instances` INT( 4 ) NOT NULL DEFAULT '1' AFTER `value`;
```
notartom, line 30: Left over comment?

application/controllers/cron/Keywords_stats.php
```diff
@@ -0,0 +1,35 @@
...
+		
+		$sql = '
+			UPDATE keywords
+			JOIN
+			(SELECT pk.keyword_id as keyword_id, COUNT(pk.project_id) AS count
+        		FROM project_keywords pk
+        		GROUP BY 1
+			) as sub
+			ON keywords.id = sub.keyword_id
+			SET keywords.instances = sub.count ';
```
notartom, line 23: You're even using `count` here yourself ;)

application/models/Project_model.php
```diff
@@ -419,6 +493,7 @@ public function get_authors_by_project($project_id, $type= 'author')
         return false;
     }
 
+
```
notartom, line 496: Extra whitespace?

application/models/Project_model.php
```diff
@@ -363,6 +363,80 @@ function get_frozen_projects()
...
+        $offset = 0 + $params['offset'];
+        $limit = 0 + $params['limit'];      
+        
+        $sql = 'select p.id, p.title_prefix, p.title, COALESCE(p.date_catalog, "2001-01-01" ) as safe_release_date, 
+        p.url_librivox, p.status, p.coverart_thumbnail, 	        p.zip_url, p.zip_size, l.two_letter_code, l.language, p.url_forum, p.project_type
+	FROM keywords k
+	JOIN project_keywords pk ON (k.id = pk.keyword_id)
+	JOIN projects p on (pk.project_id = p.id)
+	JOIN languages l ON (l.id = p.language_id) ';
+	$sql .= 'WHERE k.id = ' . $params['keywords_id'] . ' ';
```
notartom, line 395: I think this is probably safe because nothing in `$params` actually comes from user input (see two commits after this one, where this method is being called), but I always get very nervous when we do string concatenation of variables into SQL statements. It might be safe _now_, but to ascertain that we have to check every instance of this being called. We have enough SQL injection as is, could we rewrite this as an entirely parametrized query?

notartom (changes requested):
This is pretty impressive :) I only looked over the first few patches and then ran out of brain power, but I think there's enough here to post.

application/migrations/20240516002600_add_keywords_instances_field.php
peterjdann: Agreed

application/models/Project_keyword_model.php
peterjdann: Struggled with your white space issues for about 90 minutes today. Did discover I could get an add-on for gedit (my coding editor) that would show white space, but even then still struggled to figure out what git diff is showing up. I don't think I'm cut out to be a white space detective somehow. I've tried my hardest, but may not have succeeded.

application/controllers/cron/Keywords_stats.php
peterjdann: Probably. Have removed.

application/controllers/cron/Keywords_stats.php
peterjdann: You're quite right.

application/models/Project_model.php
peterjdann: Not my forte, tracking these down! I've made an effort. Have possibly been able to stomp on this one now.

peterjdann:
Was struggling today to get on top of the changes to the Migration functions and all the different CONFLICT messages git was throwing at me about this while I was doing multiple rebases and then trying to clean up my commit history. Looks like I've managed to leave things in a screwed up state with respect to that Migration change. I can't do any more on this tonight, but will look at it again tomorrow.

peterjdann:
I attempted two pushes around 14 hours ago. The first failed because I had left some glitchy code stuff in Migration.php. After I cleaned that up, I tried again. The second push/deployment has failed at a website health check. I'm not sure I know how to troubleshoot that any more from where I'm sitting, so would appreciate advice on how best to proceed from here.

peterjdann:
> > The second push/deployment has failed at a website health check. I'm not sure I know how to troubleshoot that from where I'm sitting. Would appreciate advice on how best to proceed from here.
> 
> I think I see the trouble: [this first commit](https://github.com/LibriVox/librivox-catalog/pull/225/commits/d312dddc6ceea7849f9d8fbe62d13b84e4d611d4) includes the instruction files for running the two new migrations, but it does not set `$config['migration_version'] = 3` in `application/config/migration.php`, which would tell CodeIgniter to _run_ those migrations. This causes the Page controller to crash when it tries to use the new `count` column, which hasn't been created.
> 
> If you can also remove the change to `system/libraries/Migration.php` from the same commit, I think the migration is all set without that line. 😉

Thanks for your advice, which I've followed as best I can. Deployment, at least, seems to have succeeded now.

peterjdann:
It concerns me that on the list of Pull requests (https://github.com/LibriVox/librivox-catalog/pulls) this is still showing up as "Changes requested", which seems to imply that the code has been at least partly reviewed, and I have not followed up by making requested changes. This is not the case. When changes were requested more than two months ago, I addressed them pretty promptly — and have heard nothing since.

peterjdann:
> @peterjdann - It sure has been an interesting couple of months. Now that the dust is mostly settled from the Archive incident, we are looking at this again, and I think concerns are addressed to where it can go live in a far more soonish soon.
> 
> Of course, there had to be one more thing. 😨 I sent in #247 to update some URLs in our database all at once. And since that was a migration, and I took migration #2, these will need to be 3 and 4, and config/migration.php updated in tandem. If you'd like, I could try a `git rebase` on this myself. Up to you.

I have had a go at addressing this. I _think_ I understand what's involved, and have pushed through what I _think_ is needed. Please do let me know if I've got the wrong end of the stick!

peterjdann:
> Looks like the right end to me! The end result gets us right up to date with master. Only thing I see is that the migration version is now at 3, so the (newly re-numbered) number 4 won't be called. I might also suggest keeping the two migrations in separate commits - one of them is not "undoable", but the other is.

Well, I had _part_ of the stick! Have now put the two migration actions in separate commits, and have created a separate migration version setting for each ("3" then "4").

peterjdann:
Sounds ominous. Consider me “softened up” for the bad news, then.

application/migrations/004_drop_redundant_keywords_from_projects.php
```diff
@@ -0,0 +1,53 @@
...
+			"audiolivro"
+			)
+		)';
+    		$this->db->query($sql);
+        }
+
+        public function down()
+        {
+                // revert to a previous version
+                $this->migration->version(2);
```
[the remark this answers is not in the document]
peterjdann: I agree. I've removed it.

application/models/Project_model.php
```diff
@@ -363,6 +363,85 @@ function get_frozen_projects()
...
+		$sql .= 'AND p.project_type LIKE "%" '; 
+	}
+	if ( array_key_exists('project_type_description', $params) && $params['project_type_description'] == 'group')
+	{
+		$sql .= 'AND p.project_type = "collaborative" '; 
+	}
+	if ( array_key_exists('project_type_description', $params) && $params['project_type_description'] == 'solo')
+	{
+		$sql .= 'AND p.project_type = "solo" '; 
+	}
```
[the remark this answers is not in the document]
peterjdann: My implementation here was misguided. I agree that if 'group' includes dramatic works and poetry collections, weekly. or fortnightly, in other parts of our catalog, then it should do so here. I have now followed the gist of the model you pointed out to me on lines 149-160. To test that my. new version was working as (now) intended, I have displayed audiobook titled "The Pride That Comes After", selected keywords "repentance", then filtered result. I also displayed audiobook titled "Favourite Scenes From Shakespeare" selected keywords "Shakespeare", then filtered results.

application/controllers/catalog/Page.php
```diff
@@ -61,6 +61,20 @@ public function index($slug)
 		$this->data['volunteers']->bc = $this->user_model->get($this->data['project']->person_bc_id);
 		$this->data['volunteers']->mc = $this->user_model->get($this->data['project']->person_mc_id);
 		$this->data['volunteers']->pl = $this->user_model->get($this->data['project']->person_pl_id);
+	
+		
+		// **** KEYWORDS ****//
+		$MAX_KEYWORDS_TO_DISPLAY_INITIALLY = 10;
+		$this->data['project']->has_long_keywords_list = $this->_has_long_keywords_list($this->data['project']->id, $MAX_KEYWORDS_TO_DISPLAY_INITIALLY);
```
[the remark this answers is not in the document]
peterjdann: The function is called, at most, twice, and the second time only in cases where an audiobook has a longer list of keywords than we want to display. Nonetheless, I can see that's once too often (sometimes) and have implemented the change you have requested. I have also removed the commented out line of code you mentioned.

application/controllers/cron/Keywords_stats.php
```diff
@@ -0,0 +1,32 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Keywords_stats extends CI_Controller
+{
+	public function update_keywords_stats()
```
notartom, line 5: I need to remember to add a line to https://github.com/LibriVox/librivox-ansible/blob/master/roles/common/files/daily-cron to actually run this. Is daily enough?

notartom (approved):

PR #227, opened 2024-05-19
Adding reader IDs to mp3 id3 tags

quartertone:
https://forum.librivox.org/viewtopic.php?p=2330790#p2330790

### DIFF
--- a/application/libraries/Librivox_id3tag.php
+++ b/application/libraries/Librivox_id3tag.php
@@ -118,6 +118,12 @@ class Librivox_id3tag{
 
     	$title = $this->_build_chapter_title($project, $section);
 
+      $sectionreaders = [];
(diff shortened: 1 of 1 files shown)


twinkietoes-on:
This will need MAJOR checking on the staging server before making it live. Changing something in the cataloging process could have unintended consequences.

I'm not 100% sure this is wanted; just because one or two people want it doesn't mean it should be changed. Some readers might not want it.

quartertone:
I sent a new pull request reverting the previous changes, and pivoting to adding the catalog URL instead. Hopefully this will be less objectionable.

quartertone:
Ah, got it!
I'm reading up about branches and pull requests. I'll submit a clean PR once I figure it out.

PR #229, opened 2024-05-20
Add catalog URL to ID3 tags

quartertone:
This is a proposal to add the catalog URL to the ID3 tags of the mp3 files.

Discussion continued from https://github.com/LibriVox/librivox-catalog/pull/227

(I didn't realize I could just re-branch the upstream master, so I took the circuitous route of resetting my copy of the repo to prior to my commits and then forcing a push... Well, at least now I know how to do _that_.)

### DIFF
--- a/application/libraries/Librivox_id3tag.php
+++ b/application/libraries/Librivox_id3tag.php
@@ -127,7 +127,7 @@ class Librivox_id3tag{
 		    'genre'   => array('speech'),
 		    'playtime_string' => array('0:00'),
 		    'track'   => array($track),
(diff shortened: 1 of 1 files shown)


quartertone:
Agh, sorry for the multiple pull requests. The first spacing correction fixed the indent, but I forgot about the intervening tab character, which I fixed in the second (third) pull request.

quartertone:
Right! i meant "commit". Still learning the terminology. I will send up a fresh clean pull request to keep it nice and tidy.

Thanks for your patience and the education!

PR #232, opened 2024-06-19
Improve Author autocomplete sort order

peterjdann:
This is intended to supplement the very welcome change already made to address issue #197.

The purpose of this change is to present the results shown in the autocomplete dropdown for Author fields on the Section Compiler and on the New Project Generator Template in a slightly more intelligible order for a user trying to find, say, the particular "Smith" they want from a host of other authors also named "Smith".

Because we are conducting fuzzy searches when looking to identify authors, and because some authors also have pseudonyms, it is difficult to present autodropdown results in a way that is going to satisfy all users immediately as appearing to be "logical". However, this change goes some way to addressing that requirement, and generates a rather less random-looking list than what we had previously.

Note that I have included the use of a trim() function inside the SQL used to interrogate the authors table. It appears that in the past quite a large number of authors table values were accepted which had leading spaces. Unless these are removed at the time of interrogating the database, it is impossible to get "sensible" results from the same SQL's "order by" clause (as you'll soon discover if you try removing the trim() functions from the SQL and searching on the last name "Smith").

### DIFF
--- a/application/models/Author_model.php
+++ b/application/models/Author_model.php
@@ -18,6 +18,7 @@ class Author_model extends MY_Model
 				return array();
 
 			case 'first_name':
+			// New Project Template Generator
 				$name_parts = preg_split('/ /', $term, 2, PREG_SPLIT_NO_EMPTY); // Split by first space
 
 				// Search first word as first name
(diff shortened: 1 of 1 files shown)


peterjdann:
[Before and after versions of Author dropdown with new ordering arrangement.pdf](https://github.com/user-attachments/files/15906955/Before.and.after.versions.of.Author.dropdown.with.new.ordering.arrangement.pdf)
I have attached here some example screenshots illustrating sort order appearing in Author dropdown fields on New Project Generator Template before and after the code changes included in this pull request.

peterjdann:
I believe redrun45 has come up with a much better approach than the one I've implemented here. I am therefore closing this request.

PR #238, opened 2024-08-10
Advanced Search by language: All Except English

redrun45:
Consensus: :+1: 

This adds an 'All Except English' option to the Advanced Search drop-down, and one more stanza to the Search controller to implement it.  The option is not visible anywhere except Advanced Search.

### DIFF
--- a/application/helpers/general_functions_helper.php
+++ b/application/helpers/general_functions_helper.php
@@ -116,7 +116,7 @@ function alphabet()
 }
 
 
-function full_languages_dropdown($id='', $selected_id=''  )
+function full_languages_dropdown($id='', $selected_id='', $show_all_except_english=false)
 {
     $ci = &get_instance();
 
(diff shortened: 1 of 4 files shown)


application/helpers/general_functions_helper.php
```diff
@@ -116,7 +116,7 @@ function alphabet()
 }
 
 
-function full_languages_dropdown($id='', $selected_id=''  )
+function full_languages_dropdown($id='', $selected_id='', $extra_search_options=false)
```
notartom, line 119: It's not really `$extra_search_options` though, it's more `$show_all_but_english` - maybe you have plans for this argument in the future, but for now I'd rather we be very explicit about what it's for. We can always change it later.

application/helpers/general_functions_helper.php
```diff
@@ -137,6 +137,13 @@ function full_languages_dropdown($id='', $selected_id=''  )
         $html .= '<option value="" '. $selected .' > ------ </option>';
     }
 
+    if ($extra_search_options)
+    {
+        $selected = ($selected_id == '-1') ? ' selected ': '';
```
notartom, line 142: So -1 is a magic value which, instead of being a language ID, means "all except english"? Could we then declare it as a constant, and reference that constant here instead of the literal -1?

application/libraries/Librivox_search.php
```diff
@@ -103,6 +103,24 @@ function advanced_title_search($params)
 						OR ( st.section_language_id IS NULL AND p.language_id = ' . $recorded_language . ') ) ';
 			}
 		}
+		elseif ($recorded_language === -1)
```
notartom, line 106: And then this could be way more readable, instead of being left wondering why -1 is magic ;)

notartom (changes requested):

notartom (approved):

PR #239, opened 2024-08-10
Manual sorting for genres

### LINKED ISSUE #57: Split Genre list dropdowns into 2 separate dropdowns?
In the project template generator and the add_project screen, would it be possible to have one dropdown for fiction and a separate one for nonfiction?
Genre list needs some sort of drill-down display, rather than a plain list. It's so long! And when your finger slips, you lose your place in the long list. (Try setting the genre Nonfiction / Science / Chemistry and you'll see how cumbersome it is...)

Priority: 3

redrun45:
Consensus: :+1: 
Resolves #57 - if we can adjust the ordering within the existing genre drop-down, it doesn't need to be split in two.  :grin: 

Adds a new 'Sort Order' field in Genre Manager.  Changing this number overrides the otherwise-alphabetical sorting.

Also fixes two unfiled bugs:
* The `genre.lineage` column did not have a default, so new genres could not be created.
* Genre Manager allowed setting a genre's parent as... itself, or one of its sub-items.  These genres would then need to be re-created, or fixed directly from the database.

More details in each commit, and: a special thanks to @peterjdann for figuring out that migration bug.  In case this branch makes it live before yours does, I've added you as co-author for the commit that does it here.  :+1:

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -2,5 +2,5 @@
 
 $config['migration_enabled'] = TRUE;
 $config['migration_type'] = 'sequential';
-$config['migration_version'] = 0;
+$config['migration_version'] = 1;
 $config['migration_path'] = APPPATH . 'migrations/';
--- a/application/controllers/admin/Genre_manager.php
+++ b/application/controllers/admin/Genre_manager.php
@@ -17,12 +17,13 @@ class Genre_manager extends Private_Controller
 	{
 		$this->data['menu_header'] = $this->load->view('private/common/menu_header', $this->data, TRUE);
 
-		$this->data['genres'] = $this->mahana_hierarchy->get();
+		$this->data['genres'] = $this->mahana_hierarchy->get_sorted_children();
 
 		$this->data['genre_dropdown'][0] = 'Top Level';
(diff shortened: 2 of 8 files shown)


application/config/migration.php
```diff
@@ -1,6 +1,6 @@
 <?php defined('BASEPATH') OR exit('No direct script access allowed');
 
 $config['migration_enabled'] = TRUE;
-$config['migration_type'] = 'sequential';
-$config['migration_version'] = 0;
+$config['migration_type'] = 'timestamp';
```
notartom, line 4: I'm surprised the migration worked in CI, I didn't think we could switch from `sequential` to `timestamp` once we had picked one.

In any case, I'd prefer to we stick to `sequential`, if that's what we started with. I don't care about specifically _which_ it is, I care about it being consistently the same.

notartom (changes requested):

application/config/migration.php
notartom: In that case, yeah, let's leave it `sequential`, and ask @peterjdann to update his PR to use `sequential` as well.

notartom (approved):

PR #242, opened 2024-08-26
Update author stats correctly, and show plurality

redrun45:
Replaces #14 - we may as well benefit from fixing the in-progress statistic, and showing plurality for "book" vs "books".  Removing the statistics line isn't something we're sure we want.  Better to add to it instead, when we get around to that.  :wink:

### DIFF
--- a/application/controllers/cron/Project_status_stats.php
+++ b/application/controllers/cron/Project_status_stats.php
@@ -22,6 +22,14 @@ class Project_status_stats extends CI_Controller
 
 	public function author()
 	{
+		// Reset completion and in-progress counts for all authors
+		// This avoids an edge-case in the later queries: if there are no projects with the given author or status, they leave meta_complete
+		//  and meta_in_progress untouched, resulting in 'phantom' project counts.
+		$this->db->query('
+			UPDATE authors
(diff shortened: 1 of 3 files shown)


application/controllers/cron/Project_status_stats.php
```diff
@@ -22,6 +22,11 @@ public function __construct()
 
 	public function author()
 	{
+		$this->db->query('
```
notartom, line 25: I think a comment here would be necessary - I had to look at the commit message to understand _why_ this is being done.

application/helpers/previewer_helper.php
```diff
@@ -255,3 +255,9 @@ function format_playtime($seconds)
 	$hours     = ($remainder - $minutes) / 60;
 	return sprintf('%02d:%02d:%02d', $hours, $minutes, $seconds);
 }
+
```
notartom, line 258: Can we split these two into a separate commit? I could be missing something (I guess we can keep the same PR?), but it seems unrelated to the first file.

notartom (changes requested):

notartom (approved):

PR #246, opened 2024-11-06
Add cover art to the API

### LINKED ISSUE #245: Feature Request: Add cover image to API
Hi,
Thanks for releasing the LibriVox API! Would it be possible to add the cover image to the API?
Thanks!

fakerybakery:
Hi,

Thank you for making the LibriVox API available! This PR aims to add cover art support (PDF/JPG/thumbnail) to the API when `?extended=1`. I'm still working on setting up the catalog locally, so I haven't been able to fully test it yet.

Thanks!
(Closes #245)

### DIFF
--- a/application/controllers/api/Feed.php
+++ b/application/controllers/api/Feed.php
@@ -22,6 +22,7 @@ class Feed extends REST_Controller
 
 		$params['since'] = $this->get('since');
 		$params['extended'] = $this->get('extended');
+		$params['coverart'] = $this->get('coverart');
 		//$params['simple'] 	= $this->get('simple');
 		$params['genre'] = $this->get('genre');
 		$params['title'] = $this->get('title');
--- a/application/libraries/Librivox_API.php
+++ b/application/libraries/Librivox_API.php
@@ -72,6 +72,7 @@ class Librivox_API{
 		//return $params;
 
 		$extended = (empty($params['extended'])) ? 0 : 1;
+		$show_cover_art = (empty($params['coverart'])) ? 0 : 1;
 
 		$limit = (!empty($params['limit'])) ? $params['limit'] : 50;
 		$offset = (!empty($params['offset'])) ? $params['offset'] : 0;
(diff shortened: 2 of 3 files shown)


fakerybakery:
Hi,
Thanks for the details! I've installed the catalog locally and tested it with several books (including those without covers). It seems to be working properly – the `coverart_jpg`/`coverart_pdf`/`coverart_thumbnail` return the IA URL (or `null` if no cover is available). I've also squashed the changes into a single commit.
Thanks!

fakerybakery:
Thanks for the feedback! I've added a `coverart` parameter (`?coverart=1`). Please let me know if there's anything else that needs to be changed.

notartom:
> And, it appears the check has failed due to a temporary network error (can't download Wordpress). I can't clear it to run again, and I _also_ won't be available to test this version until at least next week, but at least it's seen. 😅

Not quite - WordPress.com took over ACF and renamed it, changing its URL and making the one we were using to download it invalid. There was a whole kerfuffle about it, even made its way onto Hacker News [1]. https://github.com/LibriVox/librivox-ansible/pull/33 fixes it.

[1] https://news.ycombinator.com/item?id=41821336

notartom:
@redrun45 as far as I know, the standard practice way of evolving APIs is to version them. In my OpenStack day job, we use a thing called "microversions". I believe the same overall idea applies elsewhere, even in web apps. The principle is that for every version of the API, the set of things that can be requested, as well as the possible responses, are well defined and fixed. Any chance to the API, be it in input or output, is a new version.

If we get to the point that we have enough changes to our API, we can move in that direction. I think that _for now_, what @fakerybakery has proposed is good enough.

[1] https://docs.openstack.org/api-guide/compute/microversions.html

notartom (approved):

notartom:
@fakerybakery I'm fine with this as-is. Before merging, would you be willing to add [tests](https://github.com/LibriVox/librivox-catalog/tree/master/application/tests) for your new API? We have an old code base with a lot of gaps to cover, but at least for new things we're trying to move towards having tests where reasonable, and this is the perfect candidate for a test :) If not, either @redrun45 or myself can follow up with a separate PR.

notartom (approved):

notartom:
I've attempted tests for this in https://github.com/LibriVox/librivox-catalog/pull/248

I probably got a whole bunch of stuff wrong, but I wanted to get something posted to prove it out. The main point of that PR is the DB seeder based on https://github.com/kenjis/ci-phpunit-test/blob/3.x/docs/HowToWriteTests.md#database-seeding that allows us to load up our scrubbed DB before every test.

PR #252, opened 2025-01-07
Fix zero-sized files (and some warnings)

### LINKED ISSUE #31: in progress icon hangs when adding from section compiler on validator screen
Jo: When adding sections to the validator from the Section Compiler, the in-progress icon keeps spinning even though all the files have been added. One user went away and came back 3h later and the spinner was still spinning! Firebug isn't helpful here - hasn't reported any errors when I have had it running. Workaround is to open the catalog preview page in a new tab and once all the section durations are showing, you can hit escape or reload the validator page to continue. Anything else I can do to help diagnose the problem? 

Jeff: does this happen every time, or was just an occasional reported instance? 

Jo: Not every time, but frequently. 

Jeff: ok, I doubt this has to do with file size, and is instead some function that's not completing and so not returning a value to the script. I'll do a quick look over, but we'll need to try to grab the project id, etc next time one happens so we can trace.

Priority: 2

### LINKED ISSUE #119: Improve handling of incorrect paths in validator

redrun45:
The first fixes #119, by removing the Validator code that would fetch MP3 files from remote servers.  A bad fetch (from a typo) would often result in a zero-sized file in the Validator folder.
This 'fetch' feature was handy before Uploader became prevalent, but is no longer used.  Better to remove than to fix.  Admin discussion:
https://forum.librivox.org/viewtopic.php?p=2409190#p2409190

The next fixes #31 on *dev*, where mere warnings still cause a hang.  In prod, the only cause I see for hanging is those zero-sized files we just addressed.  We can consider this fixed as well, until we hear otherwise.

Future-musings:
This does look ripe for a re-factor (now `_local_file` and `_copy_file` can be merged, for example), but I'll leave that be for now.  I wouldn't mind removing the `test_*` functions either (they don't look to automate well), but we might as well keep them in sync while they remain.

### DIFF
--- a/application/controllers/private/Validator.php
+++ b/application/controllers/private/Validator.php
@@ -324,7 +324,7 @@ class Validator extends Private_Controller
 			$this->librivox_id3tag->_update_section($section->id, array('file_name' => $file_name, 'playtime' => $playtime));
 		}
 
-		$this->ajax_output(array('message' => 'Files copied.', 'tags' => $tag_data, 'full_tags' => $fullid3_tags), TRUE);
+		$this->ajax_output(array('message' => 'Files copied.'), TRUE);
 	}
 
 	function get_file_tags()
--- a/application/controllers/private/Validator.php
+++ b/application/controllers/private/Validator.php
@@ -298,12 +298,8 @@ class Validator extends Private_Controller
 			{
 				$file_name = $this->_copy_file($local_file, $copy_to_dir);
 			}
-			else
-			{
-				$file_name = $this->_get_file($section->listen_url, $copy_to_dir);
-			}
 
-			if (!$file_name) continue;
+			if (!$local_file or !$file_name) continue;
 
 			$file_name = trim($file_name);
             $file_path = $copy_to_dir . '/' . $file_name;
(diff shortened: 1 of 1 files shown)


notartom (approved):
I love me anything that _removes_ code :)

PR #256, opened 2025-05-13
FUP to PR#231 - keywords search dropdown for advanced search

notartom:
* Remove migration and rely on cron
* Remove APCU as it's a key-value cache, and doesn't make much sense to use for a single JSON blob


notartom:
@redrun45 I actually need your approval on this - the approval configuration in GitHub is preventing me from approving my own PR - which is probably a good thing at this point? :)

PR #265, opened 2025-08-10
Use two-letter language codes in RSS feeds, where available

redrun45:
Evidently, RSS feeds are supposed to use the two-letter language codes, instead of three.  Switch to using two-letter codes where feasible:
https://forum.librivox.org/viewtopic.php?p=2479460#p2479460

A couple notes on tests.  Evidently, we had a stub RSS controller file sitting around, and our test harness preferred to send requests to that, instead of to the real one.  Also, our test harness includes little debug messages both before and after the response itself, so I had to do a little cleanup in order to validate the XML.  That validation might be a standard part of any new tests I write going forward...


application/tests/controllers/rss/Rss_test.php
```diff
@@ -0,0 +1,83 @@
...
+   * @dataProvider provider
+   */
+  public function test_index($id, $expected)
+  {
+    $response = $this->request('GET', array('rss', 'index', $id));
+    foreach ($expected as $str) {
+      $this->assertStringContainsString($str, $response);
+    }
+
+    // Ugly: our test harness inserts messages like "got an instance" and "called the actual controller", before and after the actual RSS/XML response document.  Trim them off.
```
notartom, line 15: Do we know where these come form? A quick grep of our own code base doesn't turn up anything, and searching on GitHub the contents of phpunit and ci-phpunit-test doesn't find anything here. I'd much rather we get rid of these at the source if at all possible than working around them with hacks.

(But thanks for the code comment, I would have been left wondering otherwise).

notartom (approved):

PR #270, opened 2026-02-06
Add language filter parameter to audiobooks API

Bradenb25:
## Summary

  Adds a `language` query parameter to the `/api/feed/audiobooks` endpoint, allowing
  consumers to filter audiobooks by ISO 639-2 three-letter language codes (e.g. `fre`, `deu`, `spa`).

  This leverages the existing `languages` table join in `_build_data_set` — no additional
  joins or schema changes needed.

  ## Usage

  /api/feed/audiobooks?language=fre&format=json
  /api/feed/audiobooks?language=deu&format=json&limit=10
  /api/feed/audiobooks?language=spa&format=json&extended=1

  ## Changes

  - `Feed.php` — pass `language` query parameter to API library
  - `Librivox_API.php` — filter on `l.three_letter_code` when language param is provided (case-insensitive)
  - `api.php` (docs) — document the new parameter on the API info page

  ## Motivation

  The API currently returns a `language` field in responses but provides no way to filter by it.
  Consumers building language-specific applications (e.g. language learning tools) must fetch
  all audiobooks and filter client-side, which is wasteful.


notartom:
Hey Braden - this reads a bit like it was written with the help of AI. In and of itself that's not bad, but it's my position that it should be disclosed.

Also, while I'm not opposed to these kinds of purely additive API features that don't break backwards compatibility - I'd like to understand where they're coming from and what's the use case. For example, are you the author of a project/tool/website that's consuming/processing/parsing LibriVox catalog information, and need the ability to filter by language?

Bradenb25:
Hello, yes it was done with the help of AI. I'm more than happy to change the PR description to very clear on that. Do you have a specific way you want it put?

Yes, I'm creating a language app where people can consume different kinds of content to practice the language.

notartom:
Thanks for being transparent! In that case, yes, please call out in your commit message what tool helped you with this.

In parallel, in PR#271 we're working through an unrelated bug in the API, and that'll give us functional testing for at least a tiny bit of the API. Can I ask that we merge this PR after 271, which would allow you to to rebase and add tests for this API addition to your PR?

Issue #268, opened 2026-01-12
API returns 500 error when combining title and author parameters


mtibke:
The audiobooks API endpoint returns a 500 Internal Server Error when both title and author parameters are provided together.

# Steps to Reproduce

**Works (title only):**
curl "https://librivox.org/api/feed/audiobooks?format=json&title=Adventskalender+2023"

Returns 200 with correct data

**Fails (title + author):**
curl "https://librivox.org/api/feed/audiobooks?format=json&title=Adventskalender+2023&author=Various"

Returns 500 Internal Server Error

# Suspected Cause

In application/libraries/Librivox_API.php, the function _build_author_project_id_list() returns 0 (integer) when no matching author is found:

```
return (empty($project_ids)) ? 0 : $project_ids;

This value is then passed to where_in():
$this->db->where_in('p.id', $project_id_list);
```

CodeIgniter's where_in() expects an array, not an integer, which may cause the error.

# Suggested Fix

`return (empty($project_ids)) ? [] : $project_ids;`
And add a check in _build_data_set() to skip the where_in clause if the array is empty.

redrun45:
Thank you for digging into this!  We'll need to check that it really does work as expected after that code change - for example, I'm not sure why there wouldn't be an author found - but this is a solid lead, and I for one appreciate the detail.  👍

notartom:
This looks like a good candidate for a reproducer test patch. I strongly suspect we can demonstrate the bug in a test, assert we get the 500, and then follow that up with a patch that both fixes the bug, and adjusts the test to assert the correct behaviour. It might be possible to avoid "manual" work and do this entirely in CI :)

PR #271, opened 2026-02-07
Fix combined title+author API query producing invalid SQL (#268)

notartom:
Previously, _build_author_project_id_list() and
_build_genre_project_id_list() ran their subqueries after the main
query's WHERE clauses (e.g. p.title) had already been added to
CodeIgniter's shared query builder. Because the builder is stateful
and accumulates WHERE clauses until a ->get() consumes them, clauses
referencing the 'p' alias leaked into the author/genre subqueries,
which query different tables without a 'p' alias, producing invalid
SQL. Additionally, _build_author_project_id_list() could return 0
instead of an empty array when no authors matched, which
where_in() would not handle correctly.

With this patch, both subqueries are moved to the top of
_build_data_set(), before any main query WHERE clauses are added to
the builder, so the builder is still clean when they execute. Both
helper methods now also always return a results array (never 0),
making the subsequent empty() guard and where_in() behave correctly
in all cases. Together, these changes fix both underlying root
causes: the WHERE clause leaking into the subqueries, and the
potential return of 0 being unhandled.

A regression test is included that requests the API with both title
and author params and asserts a 200 response.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/application/libraries/Librivox_API.php
+++ b/application/libraries/Librivox_API.php
@@ -98,6 +98,22 @@ class Librivox_API{
 
 		*/
 
+		// Run subqueries first, while the query builder is clean. The
+		// builder is stateful: WHERE clauses accumulate until a ->get()
+		// consumes and resets them. If we added main query clauses (e.g.
+		// p.title) before running these subqueries, they would leak into
+		// the subqueries (which query different tables without a 'p'
+		// alias), producing invalid SQL.
+		if (!empty($params['author']))
+		{
+			$author_project_id_list = $this->_build_author_project_id_list($params['author']);
+		}
+
+		if (!empty($params['genre']))
+		{
+			$genre_project_id_list = $this->_build_genre_project_id_list($params['genre']);
+		}
+
 		if ($params['id']) $this->db->where('p.id', $params['id']);
 
 		if ($params['since']) $this->db->where('UNIX_TIMESTAMP(STR_TO_DATE(p.date_catalog, "%Y-%m-%d")) >=  ', $params['since']);
@@ -116,17 +132,14 @@ class Librivox_API{
 			}
 		}
 
-
-		if (!empty($params['author']))
+		if (!empty($author_project_id_list))
 		{
-			$project_id_list = $this->_build_author_project_id_list($params['author']);
-			$this->db->where_in('p.id', $project_id_list);
+			$this->db->where_in('p.id', $author_project_id_list);
 		}
 
-		if (!empty($params['genre']))
+		if (!empty($genre_project_id_list))
 		{
-			$project_id_list = $this->_build_genre_project_id_list($params['genre']);
-			$this->db->where_in('p.id', $project_id_list);
+			$this->db->where_in('p.id', $genre_project_id_list);
 		}
 
 
@@ -311,11 +324,8 @@ class Librivox_API{
 			->result_array();
 
 		require_once(APPPATH.'libraries/Underscore.php');
-
 		$project_ids = __()->pluck($result , 'project_id');
-
-		return (empty($project_ids)) ? 0 : $project_ids;
-		//return implode(', ', $project_ids);
+		return $project_ids;
 
 	}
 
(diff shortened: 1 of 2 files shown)


notartom:
Because the GitHub PR workflow sucks and Gerrit is just better in every way, I have to write this comment to record the fact that https://github.com/LibriVox/librivox-catalog/commit/d635a926b481103008f859f7af5e78aa41baebb4 reproduces part of the issue, with the "successful" CI run at https://github.com/LibriVox/librivox-catalog/actions/runs/21786057776/job/62857605160?pr=271 - successful in the sense that asserting that the code raises an exception is True). I'll now add a patch to this PR that fixes the issue and adjusts the test.

notartom:
@redrun45 thoughts? Claude Code could be controversial here. I wanted to play around with it to see what it can do - and it's genuinely impressive as an accelerator if you know what you're doing in the first place, environmental and intellectual property concerns aside.

application/libraries/Librivox_API.php
```diff
@@ -116,17 +132,14 @@ function _build_data_set($params)
 			}
 		}
 
-
-		if (!empty($params['author']))
+		if (!empty($author_project_id_list))
```
redrun45: If it were me writing this, I hope I would be editing this existing `if` as a starting point - because then I wouldn't imagine a need to add the _new_ `if` as a totally separate thing above.  Why not put line 109 inside of this block, instead of making a new one?


If it were structured that way, it would probably also be easier to catch this little disconnect: 
What happens when we search by author, using a name that doesn't exist?
* Ideally, we would return an empty result, as soon as we know there isn't a matching author.
* On live, we go ahead and build the query, with a `WHERE` that will guarantee an empty result.
* With this PR as written, when there's no author found, we _skip_ the `WHERE`, and act as though no author had been specified.

application/libraries/Librivox_API.php
```diff
@@ -116,17 +132,14 @@ function _build_data_set($params)
...
 		{
-			$project_id_list = $this->_build_author_project_id_list($params['author']);
-			$this->db->where_in('p.id', $project_id_list);
+			$this->db->where_in('p.id', $author_project_id_list);
 		}
```
redrun45, line 138: Well, I can see the appeal... but I also have questions.  Those other concerns you mentioned would be worth talking to more folks about, but here's what I see in this PR.  I don't know whether I would have made the same mistake, but I do know there is at least one.

We have our `if` block, with two statements in it.  We find out we need to move the `_build_..._list()` call so that it's earlier in the process of building the query.  Is there a reason we can't move the `where_in()` call up there, too?  Have we checked?

If we _do_ need to have them in different places, then we don't get the benefit of seeing them all together, so we need to be very careful about the way their conditions overlap (or don't).  Is there a reason we switched this original condition from checking whether the _author search parameter_ is empty, to checking whether the _list of authors matching the search_ is empty?

On live, if we search by author name, and that author doesn't exist, then we will get no results.  With this PR as written, we would instead get results from _any_ author, as if we were not searching by author name at all.

notartom: To be clear, I didn't just tell Claude "fix the bug" and push the PR. There were iterations and guidance along the way, with me trying to understand and double check what it did. So any mistakes are my responsibility, not AI's ;)

In terms of having them in separate places... I genuinely don't know if there's a better solution. This is such messy spaghetti code that it's hard to follow all the branches and the `WHERE`s and the `LIKE`s. Assuming we do need them in separate places, you raise a good point that the conditions have been subtly changed with the current proposal. If there's an author clause in the search but it comes up empty, the expected behaviour is to have the entire search turn up nothing, because there's no author match. With this proposal, we'd still run all the other clauses, and treat the search as if the author clause didn't exist.

I'll need some time to fix this, mostly because I want the tests to cover all the cases...

redrun45: Understood - and all the kudos for doing more test cases!  Often, that takes far longer than the actual code change...

I suppose there's already enough difficulty in guessing an author's thought processes, that we really should assume nothing.
If it were me writing this, I'd have defaulted to keeping things together if I could.  If someone sent me this PR without any explanation, I might _assume_ they did too, but there was a reason to deviate from that.  And maybe that wouldn't be a safe assumption, AI-assistance or no.  🤔

redrun45: A potentially different technical avenue: these searches are essentially sub-queries - they run first, and then the main query is restricted to only the project IDs they return.

Key word being _return_, as it looks like they are both built and _executed_ before the main query begins execution.  If we're "leaking" `WHERE`s from these into the main query, that would seem to be a bug too, even if it's not causing other problems now.

Like you said... spaghetti code.  😓

PR #272, opened 2026-06-29
Fix typos, untranslated strings, and inconsistent array keys

Avicennasis:
## Summary

Fix 9 issues across 13 files:

- **application/language/catalan/auth_lang.php**: Fix copy-pasted docstring saying "English" instead of "Catalan"; translate untranslated `forgot_password_identity_not_found` string; fix mixed English/Catalan "Confirm" → "Confirma" in label
- **application/language/portuguese/auth_lang.php**: Translate untranslated `forgot_password_identity_not_found` string; fix "recuparar" → "recuperar" typo; fix untranslated "Group" → "Grupo" in edit group heading
- **application/language/*/project_launch_lang.php** (10 files): Fix misspelled array key `economics_polical_economy` → `economics_political_economy`; fix inconsistent space in array key `genre_teen_young adult` → `genre_teen_young_adult`
- **application/libraries/Librivox_simple_search.php**: Fix "psuedonym" → "pseudonym" in comment

The array key fixes are applied consistently across all 10 language files (English, French, Dutch, German, Italian, Spanish, Portuguese, Russian, Ukrainian, Chinese). No other code references these keys outside the language files.


redrun45:
Hello Léon, and thanks for taking the time to open a PR!  It does look like you had help, but perhaps that will make the following alterations easier too.  :wink: 

You are correct that the strings in those translations are not used anywhere.  I'm assuming they're being kept around because a lot of work went into creating them, so having them visible in the main branch should prevent duplication of work if/when we ever need them.  I do think that's unlikely, but I certainly wouldn't mind having the array keys be consistent, just in case!

When it comes to the translation strings themselves, I don't think we're accepting changes at this time.  If we intend to use them, I can imagine we might use whatever machine translation tools are available at that future time, but I also expect we'd have a fluent human review them.  No sense doing that twice!

Can you send this PR again with just the array key changes (and perhaps that comment typo, if you wish), but without any changes to the translated strings themselves?

